PR #9, opened 2019-11-19
Add null check for object_lock_configuration and cors_rule

### LINKED ISSUE #8: Set object locking conditionally
I'm not able to set object locking conditionally (ie. based on the value of a variable). When disabling object locking I set the input variable `object_lock_configuration` to an empty map `{}`. However that results in a type error.

```
Error: Inconsistent conditional result types

  on ../../main.tf line 79, in module "bucket":
  79:   object_lock_configuration = var.enable_object_locking ? {
  80:     object_lock_enabled = "Enabled"
  81:     rule                = {
  82:       default_retention = {
  83:         mode = var.object_retention_mode
  84:         days = 1
  85:       }
  86:     }
  87:   } : {}

The true and false result expressions must have consistent types. The given
expressions are object and object, respectively.
```

Similar issues have been reported upstream https://github.com/hashicorp/terraform/issues/22405

miguelaferreira:
This PR adds null checks for `object_lock_configuration` and `cors_rule` because these parameters are not easily overridden with empty values. To that end, with this PR it is possible to assign `null` to those parameters as a fallback in a condition.

This PR also adds a new example that presents a variant of the complete example, where all parameters (that are not of a primitive type) are set conditionally.

Closes #8.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,10 @@
 repos:
 - repo: git://github.com/antonbabenko/pre-commit-terraform
-  rev: v1.19.0
+  rev: v1.31.0
   hooks:
     - id: terraform_fmt
     - id: terraform_docs
 - repo: git://github.com/pre-commit/pre-commit-hooks
-  rev: v2.3.0
+  rev: v3.2.0
   hooks:
     - id: check-merge-conflict
--- a/README.md
+++ b/README.md
@@ -59,28 +59,38 @@ module "s3_bucket" {
 * [Cross-Region Replication](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/tree/master/examples/s3-replication) - S3 bucket with Cross-Region Replication (CRR) enabled
 
 <!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
+## Requirements
+
+No requirements.
+
+## Providers
+
+| Name | Version |
+|------|---------|
+| aws | n/a |
+
 ## Inputs
 
 | Name | Description | Type | Default | Required |
-|------|-------------|:----:|:-----:|:-----:|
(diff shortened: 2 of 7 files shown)


antonbabenko:
You can just pass an empty map if you don't `object_lock_configuration` and `cors_rule` to be set (`cors_rule = {}` or don't specify `cors_rule` at all - it is the same result).

I am striving for having it the same in all places and empty maps are already supported everywhere.

Please let me know if I am missing something here.

miguelaferreira:
The problem I faced is exactly that I cannot set those two variables to an empty map. The error I get says that terraform does not interpret them as maps but as objects, and therefore the only way I could initialise them empty is to create a structure where all the sub/maps are also empty. I don't think that example I add in this PR works with an empty map instead of a `null`.

antonbabenko:
`cors_rule = {}` works as expected in examples and I use it in some other places in some other modules also.

miguelaferreira:
@antonbabenko Sorry for the late reply! 

The issue I face, and am facing just now again, is that when using type `any` in a complex structure (any heterogeneous list or map), the it is hard to wrap this module in other modules because expressions like the one bellow throw type errors:
```hcl
module "bucket" {
  ...

  cors_rule = var.cors_rule != null ? var.cors_rule : {}
}
```

The errors basically say that the left- and right-hand sides of the condition do not have the same type. So my workarround is to write it like this:
```hcl
  cors_rule = var.cors_rule != null ? var.cors_rule : null
```

And that ☝️  requires a null check on the variable. 

I hope this helps to understand the issue.

miguelaferreira:
After some more testing I think an approach using try to mask the null value errors works best. I'm refactoring my branch and will push a new implementation.

bryantbiggs:
I don't quite follow the point of this PR. if you are using this module within another module, why can't you just pass in the values as variables like:

```hcl
module "bucket" {
  ...

  cors_rule = var.cors_rule
}
```

personally, I don't think this change is warranted

miguelaferreira:
Copy that 👍

PR #12, opened 2019-11-22
Fix for bucket policy count when value is not computed

### LINKED ISSUE #11: BUG: Invalid count argument
This change https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/pull/10 broke my current use of this module.

```
  on .terraform/modules/example/main.tf line 221, in resource "aws_s3_bucket_policy" "this":
 221:   count = var.create_bucket && (var.attach_elb_log_delivery_policy || var.policy != null) ? 1 : 0

The "count" value depends on resource attributes that cannot be determined
until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the
resources that the count depends on.

make: *** [plan-without-prep] Error 1
```

I believe this is because I've passing the policy via data aws_iam_policy_document.

```
module "bucket" {
  source        = "git@github.com:terraform-aws-modules/terraform-aws-s3-bucket.git?ref=v1.2.0"
  create_bucket = var.enable ? true : false

  bucket = local.bucket_name
  region = var.default_region
  policy = element(
    concat(
      data.aws_iam_policy_document.this.*.json,
      [""],
    ),
    0,
  )
}
```

antonbabenko:
Fixed #11

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ These features of S3 bucket configurations are supported:
 - server-side encryption
 - object locking
 - Cross-Region Replication (CRR)
+- ELB log delivery bucket policy
 
 ## Terraform versions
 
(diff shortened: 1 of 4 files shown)


antonbabenko:
[v1.4.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.4.0) has been released.

PR #16, opened 2020-02-27
Allow to overwrite the provider (region) when creating the bucket.

gabops:
# Description

At the moment this module does not support to overwrite the provider. This is required in scenarios where your provider configuration is defined in one region (p.g eu-west-1) and you need to create a bucket on another region (p.g us-east-1).

See https://github.com/terraform-providers/terraform-provider-aws/issues/5999 and https://www.terraform.io/docs/providers/aws/r/s3_bucket.html


AyliD:
Why wasn't this PR merged? @gabops

antonbabenko:
@AyliD There is already a better way to do the same in Terraform - https://www.terraform.io/docs/language/meta-arguments/module-providers.html#default-behavior-inherit-default-providers

```terraform
module "s3" {
  source = "..."
  
  provider = {
    aws = aws.eu-west-1
  }
}
```

Issue #7, opened 2019-11-07
[question] private is not real private access


ozbillwang:
Apply the s3 bucket change with this module with option

``` 
    acl    = "private"
```
But the real bucket's access status is `Objects can be public`, which I want to set the bucket with private as `Bucket and objects not public`

What option should I go with?

reference:

https://docs.aws.amazon.com/AmazonS3/latest/user-guide/block-public-access.html

Viewing Access Status

The list buckets view shows whether your bucket is publicly accessible. Amazon S3 labels the permissions for a bucket as follows:

**Public** – Everyone has access to one or more of the following: List objects, Write objects, Read and write permissions.

**Objects can be public** – The bucket is not public, but anyone with the appropriate permissions can grant public access to objects.

**Buckets and objects not public** – The bucket and objects do not have any public access.

**Only authorized users of this account** – Access is isolated to IAM users and roles in this account and AWS service principals because there is a policy that grants public access.

go through this url, seems no one is suitable

https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl

ozbillwang:
https://aws.amazon.com/blogs/aws/amazon-s3-block-public-access-another-layer-of-protection-for-your-accounts-and-buckets/

https://www.terraform.io/docs/providers/aws/r/s3_bucket_public_access_block.html

antonbabenko:
Hi, Bill!

Yes, you are right. This module has to implement `aws_s3_bucket_public_access_block` resource as well.

I will reopen this issue as a reminder for myself when I work on this.

antonbabenko:
Hi all,

[v1.6.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.6.0) has been just released. It has support for blocking public access, see an example:

https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/blob/0b4680a22110689b12cbc61cb016a5cad83feecc/examples/complete/main.tf#L178-L182

hanswesterbeek:
Ok so one must add:

```terraform
resource "aws_s3_bucket_public_access_block" "all_private" {
  bucket = aws_s3_bucket.nicebucket.id

  block_public_acls       = true
  block_public_policy     = true
  ignore_public_acls      = true
  restrict_public_buckets = true
}
```

Issue #17, opened 2020-02-28
Add aws_s3_bucket_public_access_block to this module


andrii29:
Hello,
please add
https://www.terraform.io/docs/providers/aws/r/s3_bucket_public_access_block.html
resource to this module

Regards,
Andrii

antonbabenko:
Hi @andrii29 ,

[v1.6.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.6.0) has been just released. It has support for blocking public access, see an example:

https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/blob/0b4680a22110689b12cbc61cb016a5cad83feecc/examples/complete/main.tf#L178-L182

andrii29:
Hello,

please add this values to outputs

Regards,
Andrii

PR #18, opened 2020-03-02
Aws s3 bucket public access block

serhatcetinkaya:
Prior to this change, the module didn't have support for blocking public access check [here](https://docs.aws.amazon.com/AmazonS3/latest/dev/access-control-block-public-access.html) and [here](https://www.terraform.io/docs/providers/aws/r/s3_bucket_public_access_block.html).

All the blocking defaults to `true` and can be adjusted with variables. Hopefully this PR resolves issues #7 and #17 .

New version can be tested with following example:
```
module "s3_bucket" {
  source = "github.com/serhatcetinkaya/terraform-aws-s3-bucket?ref=aws_s3_bucket_public_access_block"

  bucket = "my-s3-bucket-test-blocking-public-access"
  acl    = "private"

  versioning = {
    enabled = true
  }

}
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,10 @@
 repos:
 - repo: git://github.com/antonbabenko/pre-commit-terraform
-  rev: v1.21.0
+  rev: v1.27.0
   hooks:
     - id: terraform_fmt
     - id: terraform_docs
 - repo: git://github.com/pre-commit/pre-commit-hooks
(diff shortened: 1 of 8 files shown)


antonbabenko:
Thank you!

[v1.6.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.6.0) has been just released.

I updated the code a bit and set defaults to `false` (as AWS&Terraform AWS provider does).

PR #29, opened 2020-05-22
fixes README - module source path outdated

camilosantana:
Signed-off-by: camilo santana <camilo.santana@procore.com>

## Description

update example code to include new path to module

## Motivation and Context

fixes `Error: Module not found` in Terraform Cloud when using the examples provided in the README.

## Breaking Changes

no breaking changes known. this is a README.md update.

## How Has This Been Tested?

before

```hcl
Error: Module not found
The module address "terraform-aws-modules/s3-bucket/aws?ref=v1.6.0" could not
be resolved.
```

after

```hcl
Terraform will perform the following actions:
  # module.s3_bucket.aws_s3_bucket.this[0] will be created
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Only Terraform 0.12 is supported.
 
 ```hcl
 module "s3_bucket" {
-  source = "terraform-aws-modules/s3-bucket/aws"
+  source = "github.com/terraform-aws-modules/terraform-aws-s3-bucket"
 
   bucket = "my-s3-bucket"
   acl    = "private"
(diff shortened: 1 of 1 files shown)


antonbabenko:
You should not add `?ref=` in the source when pointing to a module from the Terraform Registry.

Here is more information - https://www.terraform.io/docs/modules/sources.html#terraform-registry

camilosantana:
> You should not add `?ref=` in the source when pointing to a module from the Terraform Registry.

sure but did you look at the diff?

image

you can suggest an edit or close the PR and keep a bad path in the documentation.

antonbabenko:
Yes, I looked into it, and as I pointed to the official documentation this is valid schema for module source.

Give it a try, or look into right side here where it says "Provision Instructions" - https://registry.terraform.io/modules/terraform-aws-modules/s3-bucket/aws/1.6.0

PR #30, opened 2020-05-23
feat: Optionally allow a list of maps for cors_rule

liamdon:
## Description
Allow a list of maps as well as a map for the cors_rule variable. This matches the behavior of the Terraform's `aws_s3_bucket`.

## Motivation and Context
Not all CORS behavior can be expressed in a single rule. For example, you might want to allow GET requests from one origin but while allowing GET and PUT requests from a different one. WIth multiple rules, you can express this as follows:

```
    cors_rule = [
      {
        allowed_methods = ["GET"]
        allowed_origins = ["https://*.example.com"]
      },
      {
        allowed_methods = ["GET", "PUT"]
        allowed_origins = ["https://admin.example.com"]
      }
    ]
```

The Terraform provider allows multiple `cors_rule` blocks, and this is useful behavior.

## Breaking Changes
There are no breaking changes.

## How Has This Been Tested?
I have used this in my own module to allow multiple `cors_rule` entries, and have deployed these changes.

In addition, I have manually tested the following cases using `terraform plan` but not `terraform apply`:  

**Current behavior**
* cors_rule = map, all cors keys
* cors_rule = map, without some optional keys
* cors_rule = {}
* cors_rule = null

**New behavior**
* cors_rule = [map], all cors keys
* cors_rule = [map, map], all cors keys
* cors_rule = [map, map], without some optional keys for second map (different structural types)
* cors_rule = [map, null]
* cors_rule = [map, {}]

**Failure cases**
* The above will fail correctly when a required field is missing from any map


antonbabenko:
Thanks!

This has been just fixed in [v1.11.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.11.0).

PR #34, opened 2020-06-12
feat: Added attach_public_policy as conditional switch

Promaethius:
## Description
Add a bool which controls the creation of resource "aws_s3_bucket_public_access_block"

## Motivation and Context
We have an environment where S3 buckets are being created under a role without `PutBucketPolicy`.

## Breaking Changes
No, by default the behavior will remain the same.

## How Has This Been Tested?
Tested against a role without `PutBucketPolicy`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: git://github.com/antonbabenko/pre-commit-terraform
-    rev: v1.30.0
+    rev: v1.31.0
     hooks:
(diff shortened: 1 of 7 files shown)


antonbabenko:
Thanks, @Promaethius !

[v1.8.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.8.0) has been just released.

Promaethius:
That's a heck of a turnaround time! Thank you!

PR #35, opened 2020-06-18
fix: Using required arguments instead of lookup in replication_configuration

karancode:
## Description

For required arguments instead of doing a lookup, variables are fixed. I fixed such arguments which are required instead of lookup.

## Motivation and Context

Makes it easier to not miss these variables to set. Also, keeps the module uniform as other variables are set like this.

## Breaking Changes

None

## How Has This Been Tested?

Plan against the [example/s3-replication](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/tree/master/examples/s3-replication).

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -118,13 +118,13 @@ resource "aws_s3_bucket" "this" {
           id       = lookup(rules.value, "id", null)
           priority = lookup(rules.value, "priority", null)
           prefix   = lookup(rules.value, "prefix", null)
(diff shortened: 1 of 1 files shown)


antonbabenko:
Thanks a lot, @karancode !

[v1.12.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.12.0) has been just released.

PR #36, opened 2020-07-01
chore: Allow 0.13

pselle:
## Description
Update the required version block to allow 0.13 usage

## Motivation and Context
This module cannot be used with the 0.13 versions of terraform (beta, or the upcoming GA), this modifies the version so that it can at least be used. However, there are other changes that might be desirable, such as deprecating the `create_bucket` argument now that module count will be available.

## How Has This Been Tested?
Forked this repo, changed the file, and ran through loading the module, changing count, and changing for_each.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ module "s3_bucket" {
 
 | Name | Version |
 |------|---------|
-| terraform | ~> 0.12.6 |
+| terraform | >= 0.12.6, < 0.14 |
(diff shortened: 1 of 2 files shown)


antonbabenko:
Thanks, Pam!

[v1.9.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.9.0) has been just released.

PR #38, opened 2020-08-03
feat: Remove region parameter for 3.0 aws provider

jb-abbadie:
The 3.0 aws provider does not allow a region parameter for the `aws_s3_bucket` resource.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,13 +85,13 @@ module "s3_bucket" {
 | Name | Version |
 |------|---------|
 | terraform | >= 0.12.6, < 0.14 |
-| aws | ~> 2.35 |
+| aws | >= 3.0, < 4.0 |
 
 ## Providers
 
 | Name | Version |
 |------|---------|
-| aws | ~> 2.35 |
+| aws | >= 3.0, < 4.0 |
 
 ## Inputs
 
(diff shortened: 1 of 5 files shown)


bryantbiggs:
thanks for the PR @jb-abbadie - with the changes in the 3.0 AWS provider we will have to take a look at how we handle this transition as not everyone will be making the jump to 3.0. it is a major change so we will need some time to figure out the best approach

chris-mackubin:
In the short run, support both with a separate branch and version tagging.   Similar to what terraform did with v11 to v12.  Allows those of us that are using latest aws provider to just version stamp your module and not have to pull repo/build...

antonbabenko:
Thanks @jb-abbadie !

I think there is little sense to support both branches for AWS provider v2 and v3 simultaneously, so this PR requires AWS provider v3, and all upcoming features (other PRs) will have to be compatible with that, too.

[v1.10.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.10.0) has been just released.

If users want to use AWS provider v2 they will have to pin version of this module to be v1.9.0 or older.

kimmoahokas:
Hey,

First of all, thanks for the module, it has been great.

But, a small feedback about this change: Many of us have pinned this module to version 1.x and aws provider to version 2.x. Now, my `terraform init` just failed as it tried to install 1.10 of this module, which is no longer compatible with aws 2.x. It would be nice if this was version 2.0 or something :)

antonbabenko:
In theory, yes, it is possible, but after reading [CHANGELOG](https://github.com/terraform-providers/terraform-provider-aws/blob/master/CHANGELOG.md) carefully I don't see anything that is breaking in version 3 except `region` which is fixed in this PR.

I recommend locking module version stricter to have it working as expected (for eg, `version = "1.9.0"`).

kimmoahokas:
> I recommend locking module version stricter to have it working as expected (for eg, `version = "1.9.0"`).

Yep, that's what we did now and everything is working again. I guess we just had a different expectation about versioning :)

antonbabenko:
Good that it works! I support you, and you are right about versioning. I would say that this time I was just lazier than usual to not make a major release. Saved me a few minutes of my life 👍

PR #40, opened 2020-08-05
feat: support a list of rules instead of a single rule

exocom:
## Description

Modify CORS to support multiple rules instead of a single rule
## Motivation and Context

When importing an existing bucket the module was attempting to remove all but one of the rules.
## Breaking Changes

Yes

Object to Array. It is possible to be backwards compatible, however, since the syntax is as simple as adding brackets there is little reason to add additional code.

## How Has This Been Tested?

Applying on an existing resource

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -76,8 +76,8 @@ variable "website" {
   default     = {}
 }
 
-variable "cors_rules" {
-  description = "List containing rules for Cross-Origin Resource Sharing."
+variable "cors_rule" {
+  description = "List of maps containing rules for Cross-Origin Resource Sharing."
   type        = list(any)
   default     = []
 }
--- a/variables.tf
+++ b/variables.tf
@@ -82,10 +82,10 @@ variable "website" {
   default     = {}
 }
 
-variable "cors_rule" {
-  description = "Map containing a rule of Cross-Origin Resource Sharing."
-  type        = any # should be `map`, but it produces an error "all map elements must have the same type"
-  default     = {}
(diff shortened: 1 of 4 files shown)


variables.tf
```diff
@@ -82,10 +82,16 @@ variable "website" {
   default     = {}
 }
 
-variable "cors_rule" {
-  description = "Map containing a rule of Cross-Origin Resource Sharing."
-  type        = any # should be `map`, but it produces an error "all map elements must have the same type"
-  default     = {}
+variable "cors_rules" {
+  description = "List containing rules for Cross-Origin Resource Sharing."
+  type        = list(object({
```
bryantbiggs, line 87: lets use
```hcl
type = list(any)
```

instead. if we spell out each attribute like below, this breaks the optional arguments and becomes very strict

exocom: @bryantbiggs changes have been made.

Personally I'm a fan of the strict, however, I also understand the desire for optional arguments :)

exocom:
@bryantbiggs any chance we can get this merged in?

bryantbiggs:
@antonbabenko :+1:

antonbabenko:
I renamed variable to be the same as before - `cors_rule`. This way users should wrap value to make it `list`.

antonbabenko:
Thanks @exocom for this PR!

[v1.11.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.11.0) has been just released.

PR #44, opened 2020-09-14
feat: add bucket acl policy grants

### LINKED ISSUE #20: Feature Request - S3 ACL Policy Grant
Hi there,

We have a requirement to implement Bucket ACLs on a few buckets in S3 and have been using this module for other buckets we have created, so we'd like to keep some consistency if possible. It looks like this module doesn't support grants (https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#grant).

Is this not supported by design? I know there's some weirdness with `acl` vs `grant` using the S3 Resources. I have some time over the weekend and might be able to work on this, but if it's purposefully not supported I wouldn't want to waste my time

bryantbiggs:
## Description
- add ability to provide `grant` statement blocks
- update pre-commit hooks to latest
- update complete example to add cloudfront log bucket example using `var.grant` added here

closes #20 

## Motivation and Context
- `grant` is useful in scenarios when setting up access for CloudFront to write logs to your bucket or with cross account access to write files to a bucket you own

## Breaking Changes
- no

## How Has This Been Tested?
- via updating and deploying complete example

### DIFF
--- a/examples/complete/main.tf
+++ b/examples/complete/main.tf
@@ -204,7 +204,7 @@ module "s3_bucket" {
     }
   }
 
-  // S3 bucket-level Public Access Block configuration
+  # S3 bucket-level Public Access Block configuration
   block_public_acls       = true
   block_public_policy     = true
   ignore_public_acls      = true
--- a/examples/complete/main.tf
+++ b/examples/complete/main.tf
@@ -2,6 +2,8 @@ locals {
   bucket_name = "s3-bucket-${random_pet.this.id}"
 }
 
+data "aws_canonical_user_id" "current" {}
+
 resource "random_pet" "this" {
   length = 2
 }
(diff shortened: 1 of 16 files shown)


examples/complete/main.tf
```diff
@@ -55,6 +57,25 @@ module "log_bucket" {
   attach_elb_log_delivery_policy = true
 }
 
+module "cloudfront_log_bucket" {
+  source = "../../"
+
+  bucket = "cloudfront-logs-${random_pet.this.id}"
+  acl    = null # conflicts with default of `acl = "private"` so set to null to use grants
```
bryantbiggs, line 64: this does work; since `acl` and `grant` conflict (and you will get a message stating as much unless you add this). the alternative would be to change the default for `acl` to be `""` or `null`, but that would be a breaking change so this seemed like a better compromise. maybe there is a place to better document this? open to suggestions

bryantbiggs:
@antonbabenko let me know if you have any thoughts on the proposed changes

antonbabenko:
Thanks, @bryantbiggs !

[v1.13.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.13.0) has been just released.

Issue #59, opened 2020-12-09
ELB Log Delivery Policy does not work in Gov Cloud


duganth-va:
Looks like the resource ARN would need to be changed for gov cloud to `aws-us-gov`. Any reason the module creates the arn instead of using the ARN from the resource output?

antonbabenko:
[v1.17.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.17.0) has been just released.

PR #60, opened 2020-12-09
Change ELB Log Delivery Policy to use ARN output

duganth-va:
## Description
The ELB log delivery policy creates an ARN as opposed to using the S3 resources ARN output. These results in the policy not working in gov cloud. 
https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/issues/59

## How Has This Been Tested?
I test this change in gov cloud.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -257,7 +257,7 @@ data "aws_iam_policy_document" "elb_log_delivery" {
     ]
 
     resources = [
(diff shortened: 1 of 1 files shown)


antonbabenko:
Thank you for the fix!

[v1.17.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.17.0) has been just released.

PR #69, opened 2021-02-23
chore: Run pre-commit hooks

antonbabenko:


antonbabenko:
@bryantbiggs What do you think we can do to be able to pass checks in `examples`?

I am personally in favor of bumping versions requirements of many modules to be 0.13 but the code actually works with 0.12 too, so it does not feel completely ok to bump it without real need. What do you think?

bryantbiggs:
I've been trying to upgrade to 0.12.26 which has a lot of experimental support without making the leap to 0.13.x and its worked for most cases - however, maybe we are trying to support too much at once. I have this PR where the examples fail due to a 0.13.x feature but the module works fine on 0.12.26 https://github.com/terraform-aws-modules/terraform-aws-route53/pull/31

antonbabenko:
Terraform 0.12.26 seems to be the easiest solution to make big dependencies/users happy. Resolved in #70

PR #75, opened 2021-03-22
fix: ACL value can be null even when using Terragrunt

antonbabenko:


antonbabenko:
[v1.22.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v1.22.0) has been just released.

PR #101, opened 2021-07-15
feat: object_ownership

carnei-ro:
## Description

Allow configuring "Object Ownership"

## How Has This Been Tested?


utako:
would love to see this get merged soon.

hiloboy0119:
I'm currently doing this outside of the module and trying to update Ownership Controls and Public Access Policy at the same time results in the error 

```A conflicting conditional operation is currently in progress against this resource. Please try again.```

You will need to set up a dependency to keep them from running at the same time.

PR #104, opened 2021-08-04
fix: Always send `filter` map in replication config

smrz2001:
## Description
Send an empty `filter` map if no (or empty) `filter` was specified in `replication_configuration`.

(I'm not a Terraform expert so I'm not sure if the way I've done it is the best 😅)

## Motivation and Context
In the absence of at least an empty `filter` map if no filter was specified in a rule, `terraform apply` gives the following error for multi-destination S3 replication rules. See [this](https://github.com/hashicorp/terraform-provider-aws/issues/16546) `terraform-provider-aws` issue for additional context.

`Number of distinct destination bucket ARNs cannot exceed 1`

## Breaking Changes
Does not break backwards compatibility with the current major version.

## How Has This Been Tested?
- [x]  I have tested and validated these changes using one or more of the provided `examples/*` projects.
- Tested using the `S3 bucket with Cross-Region Replication (CRR) enabled` example.
- Tweaked the example to replicate `s3_bucket` to two replica buckets in the same region (one replication rule without a filter specified).
- Observed the `Number of distinct destination bucket ARNs cannot exceed 1` error with unmodified code and successful application with modifications.

PR #105, opened 2021-08-13
fix: Always send `filter` map in replication config

antonbabenko:
Completely based on #104

Fixes #104


antonbabenko:
[v2.7.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/releases/v2.7.0) has been just released.

PR #107, opened 2021-08-26
fix: fixed conflicting conditional bucket public access block

brunoxd13:
## Description

Add dependency block on creation of `aws_s3_bucket_public_access_block`.

## Motivation and Context

Fix the following error: 
`Error: error creating public access block policy for S3 bucket: OperationAborted: A conflicting conditional operation is currently in progress against this resource. Please try again.`

And relates to issue: https://github.com/hashicorp/terraform-provider-aws/issues/7628

## How Has This Been Tested?
- [X] I have tested and validated these changes using one or more of the provided `examples/*` projects

I have deployed all example in my personal AWS account to check if have some breaking change, and it's ok.


bryantbiggs:
it looks like this is a bug thats fixed in the upstream AWS provider so we should be ok to close this now, correct? 

https://github.com/hashicorp/terraform-provider-aws/issues/7628 - users will need to update to v3.67.0

PR #112, opened 2021-10-15
Fix code to replication_configuration var have possibility be null

fdmsantos:
## Description
Inside replication_configuration dynamic module, i have a try in for_each expression, to evaluate the expression and returns 0 if the expression fails. The expression will fail if the variable is null

## Motivation and Context
I want add replication based in variable. So if variable is true, add the configuration, if it's false put the replication null. The problem is the for_each expression don't accept null.

`╷
│ Error: Invalid function argument
│ 
│   on ../../modules/s3/main.tf line 126, in resource "aws_s3_bucket" "this":
│  126:     for_each = length(keys(var.replication_configuration)) == 0 ? [] : [var.replication_configuration]
│     ├────────────────
│     │ var.replication_configuration is null
│ 
│ Invalid value for "inputMap" parameter: argument must not be null.
╵
Error running command: terraform plan -input=false
`

if i put the {} value if i don't want replication i will get the following error:

` 
The true and false result expressions must have consistent types. The given
expressions are object and object, respectively.
`

With this, i can put null if i don't want replication. I can use this configuration:

`
s3_crr_config = var.enable ?  {
    role = <role_arn>
    rules = [
      {
        id     = "replicate-all"
        status = "Enabled"
        delete_marker_replication_status = "Enabled"
        filter = {}
        destination = {
          bucket        = <bucket>
          storage_class = "GLACIER"
          account_id = <account>
          access_control_translation = {
            owner = "Destination"
          }
        }
      },
    ]
  } : null
`

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -123,7 +123,7 @@ resource "aws_s3_bucket" "this" {
 
   # Max 1 block - replication_configuration
   dynamic "replication_configuration" {
-    for_each = length(keys(var.replication_configuration)) == 0 ? [] : [var.replication_configuration]
+    for_each = try(length(keys(var.replication_configuration)), 0) == 0 ? [] : [var.replication_configuration]
 
     content {
       role = replication_configuration.value.role


antonbabenko:
Try something like this:

```
replication_configuration = { for k in [true]: { role = "...", rules = ... } if var.enable_replication }
```

It is ugly but it can be a workaround.

berengamble:
@fdmsantos how did you get this working? @antonbabenko's suggestion doesn't work for me.  I get `Key expression is required when building an object.`

Maybe because it wrote it on his mobile 😛

antonbabenko:
There are a few ways to write this, for example:

```hcl
replication_configuration = lookup({ for k in [true]: "_" => { role = "...", rules = "..." } if var.enable_replication }, "_", {})
```

Next time I should write it on a better mobile phone, I agree. :)

berengamble:
That worked! Thanks!!

fdmsantos:
Hi,

Yes, i had to do some chances in Anton Suggestion.

Tks Again Anton for all work to terraform community.

Regards,
Fábio Santos

A quarta, 10/11/2021, 16:42, Beren Gamble ***@***.***>
escreveu:

antonbabenko:
Glad that it works (at the end) :)

Issue #86, opened 2021-04-19
Amazon S3 Bucket Policy for CloudTrail


naseemkullah:
## Is your request related to a problem? Please describe.

I'd like to easily attach the policy described [here](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/create-s3-bucket-policy-for-cloudtrail.html).

## Describe the solution you'd like.

An `attach_cloudtrail_policy` bool var that adds the following policy and attaches it to the bucket.

```hcl
data "aws_iam_policy_document" "cloudtrail_policy" {
  statement {
    sid    = "AWSCloudTrailAclCheck20150319"
    effect = "Allow"

    actions = [
      "s3:GetBucketAcl",
    ]

    resources = [
      aws_s3_bucket.this.arn,
    ]

    principals {
      type        = "Service"
      identifiers = ["cloudtrail.amazonaws.com"]
    }
  }

  statement {
    sid    = "AWSCloudTrailWrite20150319"
    effect = "Allow"

    actions = [
      "s3:PutObject",
    ]

    resources = [
      "${aws_s3_bucket.this.arn}/AWSLogs/${data.aws_caller_identity.current.account_id}/*",
    ]

    principals {
      type        = "Service"
      identifiers = ["cloudtrail.amazonaws.com"]
    }

    condition {
      test     = "StringEquals"
      variable = "s3:x-amz-acl"
      values = [
        "bucket-owner-full-control"
      ]
    }
  }
}
```

PR #127, opened 2022-01-15
feat: Add attach Cloudtrail log delivery policy option

jcolfej:
## Description
Allow to easily attach a simple policy to allow logs write from CloudTrail like described [here](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/create-s3-bucket-policy-for-cloudtrail.html).

## Motivation and Context
See https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/issues/86

## Breaking Changes
No breaking changes.

## How Has This Been Tested?
- [x] I use this in my Terraform projects

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,5 +1,5 @@
 locals {
-  attach_policy = var.attach_elb_log_delivery_policy || var.attach_lb_log_delivery_policy || var.attach_deny_insecure_transport_policy || var.attach_policy
+  attach_policy = var.attach_elb_log_delivery_policy || var.attach_lb_log_delivery_policy || var.attach_cloudtrail_log_delivery_policy || var.attach_deny_insecure_transport_policy || var.attach_policy
 }
 
 resource "aws_s3_bucket" "this" {
@@ -259,7 +259,6 @@ resource "aws_s3_bucket" "this" {
       }
     }
   }
-
 }
 
 resource "aws_s3_bucket_policy" "this" {
@@ -275,6 +274,7 @@ data "aws_iam_policy_document" "combined" {
   source_policy_documents = compact([
     var.attach_elb_log_delivery_policy ? data.aws_iam_policy_document.elb_log_delivery[0].json : "",
     var.attach_lb_log_delivery_policy ? data.aws_iam_policy_document.lb_log_delivery[0].json : "",
+    var.attach_cloudtrail_log_delivery_policy ? data.aws_iam_policy_document.cloudtrail_log_delivery[0].json : "",
     var.attach_deny_insecure_transport_policy ? data.aws_iam_policy_document.deny_insecure_transport[0].json : "",
     var.attach_policy ? var.policy : ""
   ])
@@ -309,9 +309,27 @@ data "aws_iam_policy_document" "elb_log_delivery" {
 }
 
 # ALB/NLB
-
 data "aws_iam_policy_document" "lb_log_delivery" {
   count = var.create_bucket && var.attach_lb_log_delivery_policy ? 1 : 0
+
+  statement {
+    sid = "AWSLogDeliveryAclCheck"
+
+    principals {
+      type        = "Service"
+      identifiers = ["delivery.logs.amazonaws.com"]
+    }
+
+    effect = "Allow"
+
+    actions = [
+      "s3:GetBucketAcl",
+    ]
+
+    resources = [
+      aws_s3_bucket.this[0].arn,
+    ]
+  }
 
   statement {
     sid = "AWSLogDeliveryWrite"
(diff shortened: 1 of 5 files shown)


main.tf
```diff
@@ -337,25 +355,56 @@ data "aws_iam_policy_document" "lb_log_delivery" {
...
+      type        = "Service"
+      identifiers = ["cloudtrail.amazonaws.com"]
+    }
 
+    actions = [
+      "s3:PutObject",
+    ]
+
+    resources = [
+      "${aws_s3_bucket.this[0].arn}/AWSLogs/${data.aws_caller_identity.this.account_id}/*",
```
antonbabenko, line 400: There can be `optionalPrefix` in ARN.

Also, the policy is not exactly matching the one described in the [AWS docs](https://docs.aws.amazon.com/awscloudtrail/latest/userguide/create-s3-bucket-policy-for-cloudtrail.html). e.g. missing `aws:SourceArn` in conditions.

I am not sure if it is going to work with the one you have or not. Please double-check.

antonbabenko (commented):
Looks pretty good. Please double-check that it works.

main.tf
jcolfej: No problem, I will add this ;)

jcolfej: I limit the rights on the bucket to the "most strict", as indicated on the documentation (specify path in bucket for write).
For ALB/ELB/NLB/LB logs, there is normally also a way to make the rights stricter, like [here](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html).
However, this is not what is currently implemented in the module.
Do I follow what is currently there (less strict) or do I propose to modify the others to be more strict as well?

jcolfej: So I propose a "security" improvement ;)

main.tf
```diff
@@ -337,25 +361,64 @@ data "aws_iam_policy_document" "lb_log_delivery" {
...
+    ]
+
+    condition {
+      test     = "StringEquals"
+      variable = "s3:x-amz-acl"
+      values   = ["bucket-owner-full-control"]
+    }
+
+    dynamic "condition" {
+      for_each = var.policy_cloudtrail_source_arn != "" ? [var.policy_cloudtrail_source_arn] : []
```
antonbabenko, line 414:
```suggestion
      for_each = var.policy_cloudtrail_source_arn != "" ? [true] : []
```

main.tf
```diff
@@ -337,25 +361,64 @@ data "aws_iam_policy_document" "lb_log_delivery" {
...
+
+      content {
+        test     = "StringEquals"
+        variable = "AWS:SourceArn"
+        values   = [condition.value]
```
antonbabenko, line 419:
```suggestion
        values   = [var.policy_cloudtrail_source_arn]
```

main.tf
```diff
@@ -1,7 +1,13 @@
 locals {
-  attach_policy = var.attach_elb_log_delivery_policy || var.attach_lb_log_delivery_policy || var.attach_deny_insecure_transport_policy || var.attach_policy
+  attach_policy = var.attach_elb_log_delivery_policy || var.attach_lb_log_delivery_policy || var.attach_cloudtrail_log_delivery_policy || var.attach_deny_insecure_transport_policy || var.attach_policy
+
+  policy_elb_log_prefix        = var.policy_elb_log_prefix != "" ? "${var.policy_elb_log_prefix}/" : ""
```
antonbabenko, line 4: Use this to make sure that there is one slash at the end: "${trimsuffix(var.policy_elb_log_prefix, "/")}/"

main.tf
```diff
@@ -337,25 +361,64 @@ data "aws_iam_policy_document" "lb_log_delivery" {
...
+      type        = "Service"
+      identifiers = ["cloudtrail.amazonaws.com"]
+    }
+
+    actions = [
+      "s3:PutObject",
+    ]
+
+    resources = [
+      "${aws_s3_bucket.this[0].arn}/${local.policy_cloudtrail_log_prefix}AWSLogs/${data.aws_caller_identity.this.account_id}/*",
```
antonbabenko, line 404: Could you please split this PR into two?
1. Add cloudtrail bucket policy with wider access (similar to how it is implemented now for other types of policies).
2. Tighten (as you call it, "security" improvement) existing policies via separate argument (e.g. `var.restrict_log_delivery_policies = true` (`false` by default to keep backward-compatibility))

antonbabenko (changes requested):
This PR is doing a bit too much at once. Please split it into two.

main.tf
jcolfej: Ok I do that, I will propose another PR with modifications for the other paths.

Issue #134, opened 2022-02-10
Add source_account argument to lambda_permissions in s3_notification_module


AndriySidliarskiy:
## Is your request related to a problem? Please describe.

When I enable security hub one I have one critical rule which have description - "You should ensure that access to the Lambda function is restricted to authorized principals only by using least privilege Lambda resource-based policies". Solve this problem it's  add this string to lambda permission, but in module you don't have an attribute [source_account] for lamba permissions.
{
 "StringEquals": {
  "AWS:SourceAccount": "[account_id]"
 },

## Describe the solution you'd like.

Add source_account attribute

AlKapkone:
+1
Please, fix it asap, would be grateful!

antonbabenko:
Good point. Pull request with added [`source_account`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_permission#source_account) is welcome. :)

AlKapkone:
@antonbabenko Yep, sure. I'll do it immediately.

AlKapkone:
@antonbabenko
```
git push origin source_account_lambda
ERROR: Permission to terraform-aws-modules/terraform-aws-s3-bucket.git denied to AlKapkone.
fatal: Could not read from remote repository.
```

AlKapkone:
@antonbabenko Thanks, guru! For the quick response.

PR #135, opened 2022-02-10
feat: Added source_account for lambda_permission resource

AlKapkone:
Fixes https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/issues/134


modules/notification/main.tf
```diff
@@ -64,6 +64,7 @@ resource "aws_lambda_permission" "allow" {
   qualifier           = lookup(each.value, "qualifier", null)
   principal           = "s3.amazonaws.com"
   source_arn          = local.bucket_arn
+  source_account      = var.source_account
```
antonbabenko, line 67:
```suggestion
  source_account      = lookup(each.value, "source_account", null)
```

PR #136, opened 2022-02-10
fix: Pin version to v3 due to number of breaking changes in v4

bryantbiggs:
## Description
- Pin version to v3 due to number of breaking changes in v4
- Fix mismatched value type error when using `tomap({})`

Note: running pre-comit locally fails the checks. It seems like its not respecting the `~> 3.x` and trying to validate against 4x which shows as a failure. However, going into the example directories and running `terraform init -upgrade=true && terraform plan` works as expected

## Motivation and Context
- Temporarily pin version until we can address v4 changes

## Breaking Changes
- 

## How Has This Been Tested?
	- Tested with local plans


bryantbiggs:
@antonbabenko not sure what to do here - it looks like either terraform validate or something within pre-commit is not respecting the v3.x pin. locally running an init and plan works fine though

antonbabenko:
There were several improvements (aka "potential bugs") in pre-commit-terraform during today but the fix [d362eb6](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/pull/136/commits/d362eb625ba4f340e915385d601199ab0810d535) was still required. Time to merge this one :)

Issue #137, opened 2022-02-11
Support terraform-provider-aws 4.x


mgerlach:
## Is your request related to a new offering from AWS?
No

## Is your request related to a problem? Please describe.
https://github.com/hashicorp/terraform-provider-aws 4.0.0 has breaking changes for the S3 bucket resource, see https://github.com/hashicorp/terraform-provider-aws/releases/tag/v4.0.0

## Describe the solution you'd like.
This module should work with https://github.com/hashicorp/terraform-provider-aws 4.x.x

## Describe alternatives you've considered.
We could abandon use of this module, but really don't want to ;)

## Additional context
N/A

dzzun-ops:
also faced with this problem:
`Could not retrieve the list of available versions for provider hashicorp/aws: no available releases match the given constraints ~> 3.69, 4.0.0`

kleytonhsantos:
This is the real problem:

![Screenshot from 2022-02-10 20-35-38](https://user-images.githubusercontent.com/11460558/153590689-e58629a2-a417-4d6c-a9cd-0301bf07339a.png)

To solve the problem force the version of provider to the previous version, in my case I used 3.74.1 and this was the result:
![Screenshot from 2022-02-10 20-34-52](https://user-images.githubusercontent.com/11460558/153591462-d65c79a5-a45c-4f7f-bbf7-9f1d0d67c595.png)

jdiebold:
They pulled stuff like logging and lifecycle rule out of the bucket resource into separate resources. That's why it's the values are unconfigurable.

mgerlach:
People will eventually want to use version 4.x of the provider.

antonbabenko:
Thanks for opening this issue. We are well aware of version 4.x and we will work on the support in the coming weeks.

antonbabenko:
For the time being, please use version 2.14.1 of this module where Terraform AWS provider version is locked to [`~> 3.69`](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/pull/136/files#diff-dfd5848fa77a04d0343891fe859286cc210473d10f0e62c7f99f0d5ecf1a9927R7)

baronnenate:
how do you force the use of a previous version of a module?

baronnenate:
I've set the provider version as follows:
```
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "~> 3.69"
    }
  }

[...]

PR #138, opened 2022-03-07
Upgrade S3 module to support AWS Provider 4.x

robwittman:
## Description
The release of 4.x for the AWS provider brings a large number of breaking changes for the S3 bucket configuration. This PR uses the new explicit resources for various configuration options, while limiting changes on the module interface.

## Motivation and Context
These changes are needed to support the 4.x and newer versions of the AWS provider. This module is currently locked on the 3.x branch until these are completed

Works to resolve https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/issues/137

## Breaking Changes
There are a few breaking changes around the interface itself, as well as the structure of some of the variables. 

I will update with documentation around those soon.

## How Has This Been Tested?
TBD

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -8,258 +8,15 @@ resource "aws_s3_bucket" "this" {
   bucket        = var.bucket
   bucket_prefix = var.bucket_prefix
 
-  # hack when `null` value can't be used (eg, from terragrunt, https://github.com/gruntwork-io/terragrunt/pull/1367)
-  acl = var.acl != "null" ? var.acl : null
-
-  tags                = var.tags
-  force_destroy       = var.force_destroy
-  acceleration_status = var.acceleration_status
-  request_payer       = var.request_payer
-
-  dynamic "website" {
-    for_each = length(keys(var.website)) == 0 ? [] : [var.website]
-
-    content {
-      index_document           = lookup(website.value, "index_document", null)
-      error_document           = lookup(website.value, "error_document", null)
-      redirect_all_requests_to = lookup(website.value, "redirect_all_requests_to", null)
-      routing_rules            = lookup(website.value, "routing_rules", null)
-    }
-  }
-
-  dynamic "cors_rule" {
-    for_each = try(jsondecode(var.cors_rule), var.cors_rule)
-
-    content {
-      allowed_methods = cors_rule.value.allowed_methods
-      allowed_origins = cors_rule.value.allowed_origins
-      allowed_headers = lookup(cors_rule.value, "allowed_headers", null)
-      expose_headers  = lookup(cors_rule.value, "expose_headers", null)
-      max_age_seconds = lookup(cors_rule.value, "max_age_seconds", null)
-    }
-  }
-
-  dynamic "versioning" {
-    for_each = length(keys(var.versioning)) == 0 ? [] : [var.versioning]
-
-    content {
-      enabled    = lookup(versioning.value, "enabled", null)
-      mfa_delete = lookup(versioning.value, "mfa_delete", null)
-    }
-  }
-
-  dynamic "logging" {
-    for_each = length(keys(var.logging)) == 0 ? [] : [var.logging]
-
-    content {
-      target_bucket = logging.value.target_bucket
-      target_prefix = lookup(logging.value, "target_prefix", null)
-    }
-  }
-
-  dynamic "grant" {
-    for_each = try(jsondecode(var.grant), var.grant)
-
-    content {
-      id          = lookup(grant.value, "id", null)
-      type        = grant.value.type
-      permissions = grant.value.permissions
-      uri         = lookup(grant.value, "uri", null)
-    }
-  }
-
-  dynamic "lifecycle_rule" {
-    for_each = try(jsondecode(var.lifecycle_rule), var.lifecycle_rule)
-
-    content {
-      id                                     = lookup(lifecycle_rule.value, "id", null)
-      prefix                                 = lookup(lifecycle_rule.value, "prefix", null)
-      tags                                   = lookup(lifecycle_rule.value, "tags", null)
-      abort_incomplete_multipart_upload_days = lookup(lifecycle_rule.value, "abort_incomplete_multipart_upload_days", null)
-      enabled                                = lifecycle_rule.value.enabled
-
-      # Max 1 block - expiration
-      dynamic "expiration" {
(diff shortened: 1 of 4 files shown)


examples/complete/main.tf
```diff
@@ -79,16 +79,19 @@ module "cloudfront_log_bucket" {
 
   bucket = "cloudfront-logs-${random_pet.this.id}"
   acl    = null # conflicts with default of `acl = "private"` so set to null to use grants
```
jeohist, line 81: There's no more default value, so I think it's safe to remove this now.

variables.tf
```diff
@@ -112,7 +112,7 @@ variable "logging" {
   default     = {}
 }
 
-variable "grant" {
+variable "access_control_policy" {
```
jeohist, line 115: We should update the description of var.acl too :)

main.tf
```diff
@@ -453,3 +210,402 @@ resource "aws_s3_bucket_ownership_controls" "this" {
...
 }
+
+resource "aws_s3_bucket_accelerate_configuration" "this" {
+  count  = var.acceleration_status != null ? 1 : 0
+  bucket = aws_s3_bucket.this[0].id
+  status = var.acceleration_status
+}
+
+# TODO: This should accept both ACL and Grant, exclusively
+resource "aws_s3_bucket_acl" "this" {
```
jeohist, line 221: I think we can accomplish this with this hacky bit of code (my kingdom for an exclusive OR, HashiCorp...):
```hcl
# If ACL and ACP: 0, if ACL or ACP: 1, if ACL nor ACP: 0
count = (var.acl != "null" && var.acl != null) && length(keys(try(jsondecode(var.access_control_policy), var.access_control_policy))) > 0 ? 0 : (var.acl != "null" && var.acl != null) || length(keys(try(jsondecode(var.access_control_policy), var.access_control_policy))) > 0 ? 1 : 0
```

main.tf
```diff
@@ -453,3 +210,402 @@ resource "aws_s3_bucket_ownership_controls" "this" {
...
+  dynamic "cors_rule" {
+    for_each = try(jsondecode(var.cors_rule), var.cors_rule)
+
+    content {
+      allowed_methods = cors_rule.value.allowed_methods
+      allowed_origins = cors_rule.value.allowed_origins
+      allowed_headers = lookup(cors_rule.value, "allowed_headers", null)
+      expose_headers  = lookup(cors_rule.value, "expose_headers", null)
+      max_age_seconds = lookup(cors_rule.value, "max_age_seconds", null)
+    }
```
jeohist, line 265: An optional `id` attribute is now supported, might be easy to add:
```hcl
id              = lookup(cors_rule.value, "id", null)
```

main.tf
```diff
@@ -453,3 +210,402 @@ resource "aws_s3_bucket_ownership_controls" "this" {
...
+      dynamic "filter" {
+        for_each = length(keys(lookup(rule.value, "filter", {}))) == 0 ? [{}] : []
+
+        content {}
+      }
+
+      dynamic "filter" {
+        for_each = length(keys(lookup(rule.value, "filter", {}))) == 0 ? [] : [lookup(rule.value, "filter")]
+
+        content {
```
jeohist, line 299: Maybe we can also add the newly added optional attributes?
```hcl
          object_size_greater_than = lookup(filter.value, "object_size_greater_than", null)
          object_size_less_than    = lookup(filter.value, "object_size_less_than", null)
          prefix                   = lookup(filter.value, "prefix", null)
```

jeohist (commented):
I was working on this as well, had a few additions / comments. Great to see this PR!

I'm wondering whether this might be the time to get rid of `var.create_bucket` as well? With count in modules since 0.13 and moved blocks making the transition easier, as well as this upgrade requiring lots of state file surgery anyway... Could make things a bit easier to maintain.

main.tf
robwittman: From my understanding, [aws_s3_bucket_acl](https://registry.terraform.io/providers/hashicorp%20%20/aws/latest/docs/resources/s3_bucket_acl) still supports both `acl` and `access_control_policy`, just not at the same time. Would just using something like 
```
count = (var.acl != null || var.access_control_policy != "") ? 1 : 0 
```
get the expectedc result?

main.tf
robwittman: I'd be delighted to, though I'd like to get feedback on their usage of the `and` block for predicates. For each attribute that can conditionally end up in that `and { }` segment, the check across the different filters gets more complex. Could we hoist those to an evaluation in locals maybe? Is there a straightforward way to allow passing the desired `filters` from outside the module instead?

robwittman:
> I was working on this as well, had a few additions / comments. Great to see this PR!
> 
> I'm wondering whether this might be the time to get rid of `var.create_bucket` as well? With count in modules since 0.13 and moved blocks making the transition easier, as well as this upgrade requiring lots of state file surgery anyway... Could make things a bit easier to maintain.

I don't know that I have enough background info into what use-cases `var.create_bucket` is used for. It feels like something that could be removed, but I can't back that up at all :) 

Regarding state surgery, their documentation says that since most of the new resources are PUT operations, they should mostly be no-ops, and one could safely upgrade directly to these new resources. They recommend importing where possible, but for simpler use cases, I don't think much state management will be needed

rblaine95:
I'm experimenting with this branch and it's been working really well so far!

I'm just having one issue when trying to set a lifecycle rule that automatically expires objects after 180 days and, for some reason, the `days` parameter keeps being set to `0`
```hcl
{
  id      = "Retain logs for 6 months"
  enabled = true
  expiration = {
    days = 180
  }
}
```
```hcl
  # module.example_s3_logs.aws_s3_bucket_lifecycle_configuration.this[0] will be updated in-place
  ~ resource "aws_s3_bucket_lifecycle_configuration" "this" {
        id     = "example-logs"
        # (1 unchanged attribute hidden)

      ~ rule {
            id     = "Retain logs for 6 months"
            status = "Enabled"

          ~ expiration {
              ~ days                         = 180 -> 0
                # (1 unchanged attribute hidden)
            }

            # (1 unchanged block hidden)
        }
        # (1 unchanged block hidden)
    }
```

Maybe I'm missing something stupid, but I can't seem to figure out if/what I'm doing wrong here 🤔

robwittman:
@rblaine95 If you want to pull the latest commit, there was an issue with one of the lookups that should be fixed now. Wether or not `expiration.days` was set, the `expiration` block was getting set to `for_each=[]`

rblaine95:
@robwittman just tested, working perfectly!

antonbabenko:
This issue has been resolved in version 3.0.0 :tada:

Issue #144, opened 2022-04-02
terraform apply w/ v3.0 of the module results in a loop with removal of options and then subsequent re-adding of options such as server side encryption, logging etc.


jaybocc2:
## Description

terraform apply w/ v3.0 of the module results in a loop with removal of options and then subsequent re-adding of options such as server side encryption, logging etc.

This is caused by missing the lifecycle rules that are required to use the 4.x style bucket calls in 3.75.1 
https://registry.terraform.io/providers/hashicorp/aws/3.75.1/docs/resources/s3_bucket_replication_configuration#usage-notes

Please provide a clear and concise description of the issue you are encountering, and a reproduction of your configuration (see the `examples/*` directory for references that you can copy+paste and tailor to match your configs if you are unable to copy your exact configuration). The reproduction MUST be executable by running `terraform init && terraform apply` without any further changes.

If your request is for a new feature, please use the `Feature request` template.

- [❌ ] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

- Module version [Required]:

- Terraform version:
1.1.7
- Provider version(s):
aws 3.75.1

## Reproduction Code [Required]

```
module "mybucket" {
  source = "registry.terraform.io/terraform-aws-modules/s3-bucket/aws"
  version = "~> 3.0"

  bucket = "mybucket"
  logging = {
    target_bucket = "my-access-log-bucket"
    target_prefix = "mybucket/"
  }
}
```

do multiple applies. the plan/apply will always differ either adding the logging rule via s3_bucket_logging or removing it via s3_bucket.

yes

yes

upgraded to aws 3.75.1 due to oopsie upgrade to 3.0 of module

## Expected behavior

buckets to consistently configure

## Actual behavior
enters loop with bucket configures, reconfigures

PR #145, opened 2022-04-02
fix: Add lifecycle ignore changes on s3_bucket resource to prevent configuration loop

jaybocc2:
add lifecycle ignore changes on s3_bucket resource to prevent configuration loop to fix https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/issues/144

## Description
add lifecycle ignore changes on s3_bucket resource for all new resources that are separate to prevent configuration loop

## Motivation and Context

fixes lifecycle loop

## How Has This Been Tested?

I tested it live because this is blocking my work 😞

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -32,6 +32,7 @@ resource "aws_s3_bucket" "this" {
   lifecycle {
     ignore_changes = [
       acceleration_status,
+      acl,
       grant,
       cors_rule,
       lifecycle_rule,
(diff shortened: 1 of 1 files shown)


antonbabenko:
I think I was able to reproduce the bug you described in #144, and this PR resolves it.

A message saying `Note: Objects have changed outside of Terraform` can be misleading for many users, but it is another issue with the Terraform core.

I have also added `acl` and `versioning` to `ignore_changes` just in case.

Merging and releasing now. Thank you!

antonbabenko:
This PR is included in version 3.0.1 :tada:

PR #154, opened 2022-04-27
feat: Add optional "logging.enabled" flag to module logging input

### LINKED ISSUE #153: Add enabled flag for bucket logging for more comprehensive looping

When iterating over a list of buckets to create in AWS S3 multiple module instantiations must be created with conditionals for buckets that desire logging and buckets that do not.

By adding an optional `logging.enabled` flag (defaulted to `true`) to the logging input we can reduce instantiations to a single instance which sets logging to false as necessary.

The only current alternative is to instantiate the module twice and filter a `for_each` with a conditional based on whether or not the specific bucket requires logging.

## Additional context

It'd just be really nice to have!  Reduce code complexity, make the module more DRY, etc.

jwitko:
## Description

When iterating over a list of buckets to create in AWS S3 multiple module instantiations must be created with conditionals for buckets that desire logging and buckets that do not. Adding the the parameter `logging.enabled` will make this have a more DRY approach.

Fixes #153 

## Motivation and Context

This change solves the problem where someone is looping over a list of buckets and some of them require logging and others do not. Currently this requires two separate module instantiations with `for_each` loops that filter based on some parameter added to the list of inputs.

## Breaking Changes

No breaking changes. Flag is optional and defaulted to `true` to keep existing behavior identical with no changes required from existing users.

## How Has This Been Tested?
- [X] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [X] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [X] I have executed `pre-commit run -a` on my pull request


antonbabenko:
Adding 'enabled' flag to the module is unnecessary because it is simpler not to include `logging` map completely.

There is also a [wrapper](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/tree/master/wrappers#usage-with-terraform) which can be useful for your use-case:

```hcl
module "wrapper" {
  source = "terraform-aws-modules/s3-bucket/aws//wrappers"

  items = {
    bucket1 = {
      bucket = "my-random-bucket-1"
      logging = {
        target_bucket = "my-logs"
        target_prefix = "log/"
      }
    }
    bucket2 = {
      bucket = "my-random-bucket-2"
      tags = {
        Secure = "probably"
      }
    }
  }
}
```

If you don't want to use `wrapper`, you have to create input maps using `for` loops and include/exclude `logging` values.

jwitko:
@antonbabenko I would really appreciate if you would reconsider this PR.  I finally got around to trying your wrapper method and while it is really great it still does not work with a config drive model.  One would still have to use local variables to generate the content input into the wrapper `items {}` block and if attempting to use the DAG by specifying the logging bucket as output from a separate instantiation of this module it will fail with 
```
│ The "count" value depends on resource attributes that cannot be determined
│ until apply, so Terraform cannot predict how many instances will be
│ created. To work around this, use the -target argument to first apply only
│ the resources that the count depends on.
```

antonbabenko:
Could you provide the value for `items` which does not work now?

PR #180, opened 2022-08-24
Update versions.tf

MayuriBhandari:
## How Has This Been Tested?

### DIFF
--- a/versions.tf
+++ b/versions.tf
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = ">= 4.5"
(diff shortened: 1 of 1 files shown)


antonbabenko (commented):
I wish it was so simple... If you need to use AWS provider version 3, you should use the previous major version of the module (e.g. [v2.15.0](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/blob/master/CHANGELOG.md#2150-2022-03-12)).

AWS Provider version 4 has implemented S3 bucket resources differently, so we can't just rollback to the previous version so easily.

I am going to close this PR.

PR #202, opened 2022-12-19
fix: Fixed the issue with ACLs configuration update

PavloPavlichenko:
## Description
Fix the issue with ACLs configuration update.

## Motivation and Context
In case you had `BucketOwnerEnforced `object ownership and you wanted to change it to `BucketOwnerPreferred` or `ObjectWriter` simultaneously adding ACLs - you run into the problem:
```
Error: error updating S3 bucket ACL (bucket-id): AccessControlListNotSupported: The bucket does not allow ACLs
```
In order to bypass this problem the `depends_on` block in `aws_s3_bucket_acl` resource is needed.

## Breaking Changes
No

## How Has This Been Tested?


Debayan001 (approved):
done

Debayan001:
done

lawrenceklh (approved):

erpel:
I just got the "AccessControlListNotSupported: The bucket does not allow ACLs" message while creating a bucket using code that had worked multiple times in the past.
In this case, the settings were acl = "private" and   object_ownership = "BucketOwnerEnforced".
It might be that this combination does not make sense but it seems to sometimes work and sometimes not, depending on the order of the creation. For the situation I encountered, forcing `aws_s3_bucket_acl` to be created before `aws_s3_bucket_ownership_controls` resolved it.

I assume with this change merged some settings that may have worked in the past will reliably fail - I want to try and verify this. It might be worth noting in release notes if this goes forward.

antonbabenko:
Thank you @PavloPavlichenko for this discovery and fix :)

antonbabenko:
This PR is included in version 3.8.1 :tada:

PR #231, opened 2023-05-02
Prevent accidentally creating lifecycle rule without filtering

rysi3k:
## Description
Added extra guard option to create lifecycle rule for expiration whole bucket

## Motivation and Context
In our case, we created a bucket without any filter prefix, which causes expiration of about 1,5TB of data in bucket.
We had code like snippet bellow
```hcl
module "s3_bucket" {
  source = "terraform-aws-modules/s3-bucket/aws"

  lifecycle_rule = [
    {
      enabled = true
       prefix = "prefix-to-cleanup/"
      
      expiration = {
        days = 1
      }
    }
  ]
}
```
There is a missing `filter` section wrapping `prefix`. Sadly terraform does not throw any error that config is inaccurate. That code created lifecycle rules for the whole bucket.

We wanted to be sure not to make the same mistake again and create some kind of guard before this situation. Without passing extra option, no expiration section will be created - that should be less pain than losing data ;)

Passing one extra option should not be a big deal - AWS Console have a similar checkbox to prevent creating rule for the whole bucket.

## Breaking Changes
Conditionaly/Partialy - if someone wanted to create lifecycle to cleanup whole bucket it is needed to pass extra option.

## How Has This Been Tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,67 @@ module "s3_bucket" {
 }
 ```
 
+## Expiration lifecycle rules for the whole bucket
+
+To prevent accidentally creating a lifecycle rule for the whole bucket that could remove all of your files, you must pass the `expire_all_objects_in_bucket = true` option.
+
+```hcl
+# This S3 bucket has an expiration lifecycle rule only for the specified prefix
+module "s3_bucket" {
+  source = "terraform-aws-modules/s3-bucket/aws"
+
+  lifecycle_rule = [
+    {
+      enabled = true
+
+      filter = {
+        prefix = "prefix-to-cleanup/"
+      }
+
+      expiration = {
+        days = 1
+      }
+    }
+  ]
+}
+```
+
+```hcl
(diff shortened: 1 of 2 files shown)


rysi3k:
@antonbabenko could you look at this PR? What do you think?

shbedev:
@rysi3k not all CI checks are passed. Make sure to add the convention to the start of the PR title. You can look [here](https://www.conventionalcommits.org/en/v1.0.0/) or on previous PRs to see what is a valid convention.

antonbabenko (commented):
I am not sure that this kind of enhancement will lead to safer setups. Terraform AWS modules don't provide extra guardrails because it will be tough to maintain them and reason about them. It is always up to the users to specify what they want and how they want it.

I am going to close this one.

antonbabenko:
I am not sure that this kind of enhancement will lead to safer setups. Terraform AWS modules don't provide extra guardrails because it will be very hard to maintain them and reason about them.

PR #234, opened 2023-05-25
feat: Added outputs for s3 bucket lifecycle rules and policy

dhollerbach:
## Description
This PR creates two new outputs: 1) a list of lifecycle rules applied to the s3 bucket and, 2) the policy applied to the s3 bucket. Each output is set to an empty string if its respective value is not provided.

Additionally, this PR removes the deprecated `skip_get_ec2_platforms` argument from the aws providers in all examples directories.

## Motivation and Context
While creating tests for my own S3 modules, I ran into a trouble when trying to validate lifecycle rules and policies applied to my S3 bucket. I was able to output the `data_iam_policy_document` easily enough but lifecycle rules are not available as a data source. My workaround was configuring the proper result in Go itself but it would be much better to output these values in case anyone wants to use them; for tests in my case.

Additionally, while running the pre-commit for these outputs, the validations were failing because the `skip_get_ec2_platforms` argument is not longer supported using Terraform version 1.4.5, the version used to perform all tests. To pass all validation checks, the `skip_get_ec2_platforms` argument was removed from the following examples directories:

- [examples/notification](./examples/notification)
- [examples/object](./examples/object)
- [examples/s3-inventory](./examples/s3-inventory)
- [examples/s3-replication](./examples/s3-replication)

## Breaking Changes
No.

## How Has This Been Tested?


outputs.tf
```diff
@@ -23,6 +23,16 @@ output "s3_bucket_hosted_zone_id" {
...
 }
 
+output "s3_bucket_lifecycle_configuration_rules" {
+  description = "The lifecycle rules of the bucket, if the bucket is configured with lifecycle rules. If not, this will be an empty string."
+  value       = try(jsondecode(jsonencode(aws_s3_bucket_lifecycle_configuration.this[0].rule)), "")
+}
+
+output "s3_bucket_policy" {
+  description = "The policy of the bucket, if the bucket is configured with a policy. If not, this will be an empty string."
+  value       = try(jsondecode(aws_s3_bucket_policy.this[0].policy), "")
```
antonbabenko, line 33:
```suggestion
  value       = try(aws_s3_bucket_policy.this[0].policy, "")
```

outputs.tf
```diff
@@ -23,6 +23,16 @@ output "s3_bucket_hosted_zone_id" {
   value       = try(aws_s3_bucket.this[0].hosted_zone_id, "")
 }
 
+output "s3_bucket_lifecycle_configuration_rules" {
+  description = "The lifecycle rules of the bucket, if the bucket is configured with lifecycle rules. If not, this will be an empty string."
+  value       = try(jsondecode(jsonencode(aws_s3_bucket_lifecycle_configuration.this[0].rule)), "")
```
antonbabenko, line 28:
```suggestion
  value       = try(aws_s3_bucket_lifecycle_configuration.this[0].rule, "")
```

antonbabenko (changes requested):

dhollerbach-nelnet (approved):
Looks great. I did the jsonencode / jsondecode for my own personal needs but your changes are more appropriate for a broader audience.

antonbabenko (approved):

antonbabenko:
This PR is included in version 3.11.0 :tada:

Issue #239, opened 2023-06-12
Improve SSE by specifying only one KMS key allowed


remiflament:
## Context

In the last version 3.13.0 we had the possibility to enforce SSE(s3|kms) on the bucket uploads. #238 #233

This is a great feature, and we can go further, [as the doc mentioned](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html#require-sse-kms:~:text=To%20require%20that,key%2Did%20exists.), by specifying only one key authorized to encrypt objects.

## Problem

We enforced SSE:KMS to encrypt objects. But we allow multiples keys used to do the encryption... It can be a problem when we want to perform large backup or decryption automation in a business.

## Describe the solution you'd like.

We specify a specific KMS key arn in the IAM Policy

```terraform
data "aws_iam_policy_document" "s3_put_kms_only" {
  statement {
    sid       = "RequireKMSEncryption"
    effect    = "Deny"
    actions   = ["s3:PutObject"]
    resources = ["${module.bucket.s3_bucket_arn}/*"]
    condition {
      test     = "StringNotEquals"
      variable = "s3:x-amz-server-side-encryption"
      values   = ["aws:kms"]
    }
    principals {
      type        = "*"
      identifiers = ["*"]
    }
  }

  statement {
    sid       = "RequireKMSEncryptionKeyID"
    effect    = "Deny"
    actions   = ["s3:PutObject"]
    resources = ["${module.bucket.s3_bucket_arn}/*"]
    condition {
      test     = "StringNotEquals"
      variable = "s3:x-amz-server-side-encryption-aws-kms-key-id"
      values   = ["${module.kms.key_arn}"]
    }
    principals {
      type        = "*"
      identifiers = ["*"]
    }
  }
}
```

I think I can propose the PR.
Please tell me if this enhancement makes sense to you.

antonbabenko:
I think PR would be useful for such improvement.

PR #240, opened 2023-06-17
feat: Add "deny incorrect kms key sse" bucket policy

remiflament:
## Description

Add the possibility to specify a KMS key to use as requirement for uploading files.

## Motivation and Context

See this issue #239 

## Breaking Changes

No

## How Has This Been Tested?

- [x] I have tried to upload a file without specifying the right key and got denied. Only the specified key allowed me to upload files to the bucket.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -8,7 +8,7 @@ data "aws_partition" "current" {}
 locals {
   create_bucket = var.create_bucket && var.putin_khuylo
 
-  attach_policy = var.attach_require_latest_tls_policy || var.attach_elb_log_delivery_policy || var.attach_lb_log_delivery_policy || var.attach_deny_insecure_transport_policy || var.attach_inventory_destination_policy || var.attach_deny_incorrect_encryption_headers || var.attach_deny_unencrypted_object_uploads || var.attach_policy
+  attach_policy = var.attach_require_latest_tls_policy || var.attach_elb_log_delivery_policy || var.attach_lb_log_delivery_policy || var.attach_deny_insecure_transport_policy || var.attach_inventory_destination_policy || var.attach_deny_incorrect_encryption_headers || var.attach_deny_incorrect_kms_key_sse || var.attach_deny_unencrypted_object_uploads || var.attach_policy
 
   # Variables with type `any` should be jsonencode()'d when value is coming from Terragrunt
   grants               = try(jsondecode(var.grant), var.grant)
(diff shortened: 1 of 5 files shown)


variables.tf
```diff
@@ -64,6 +64,18 @@ variable "attach_deny_incorrect_encryption_headers" {
   default     = false
 }
 
+variable "attach_deny_incorrect_kms_key_sse" {
+  description = "Controls if S3 bucket should deny incorrect kms key sse policy attached."
+  type        = bool
+  default     = false
+}
+
+variable "kms_key_sse_arn" {
```
antonbabenko, line 73:
```suggestion
variable "allowed_kms_key_arn" {
```

variables.tf
```diff
@@ -64,6 +64,18 @@ variable "attach_deny_incorrect_encryption_headers" {
...
+  description = "The kms key sse ARN."
```
antonbabenko, line 74:
```suggestion
  description = "The ARN of KMS key which should be allowed in PutObject"
```

variables.tf
```diff
@@ -64,6 +64,18 @@ variable "attach_deny_incorrect_encryption_headers" {
   default     = false
 }
 
+variable "attach_deny_incorrect_kms_key_sse" {
+  description = "Controls if S3 bucket should deny incorrect kms key sse policy attached."
```
antonbabenko, line 68:
```suggestion
  description = "Controls if S3 bucket policy should deny usage of incorrect KMS key SSE."
```

main.tf
```diff
@@ -824,6 +825,34 @@ data "aws_iam_policy_document" "deny_incorrect_encryption_headers" {
...
+
+    actions = [
+      "s3:PutObject"
+    ]
+
+    resources = [
+      "${aws_s3_bucket.this[0].arn}/*"
+    ]
+
+    principals {
```
antonbabenko, line 843: I believe, such verbose `principals` are not needed because they will be like this by default. Please remove it from here and other policies unless you think it is really needed.

antonbabenko (changes requested):
Minor comments about naming and text

main.tf
remiflament: Removing them give me this explicit error : 
`Error: putting S3 policy: MalformedPolicy: Missing required field Principal`

remiflament:
Hello @antonbabenko thanks for the feedback. I fixed the typos and text. For the "principals{}" it is mandatory you get an error when we remove them. Please could you re-check this PR?

main.tf
antonbabenko: You're right. Thanks for checking out this.

antonbabenko (approved):

PR #243, opened 2023-07-14
feat: Stop requiring `s3:ListAllMyBuckets` IAM permission unless needed (for bucket ACL)

theipster:
## Description

This skips executing the `data.aws_canonical_user_id.this` data source unless it is actually needed.

The data source is only needed when the `aws_s3_bucket_acl.this` resource needs to be created **and** the `var.owner["id"]` value isn't available.

## Motivation and Context

As per [the `data.aws_canonical_user_id` documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/canonical_user_id), this data source requires the `s3:ListAllMyBuckets` IAM permission. Note that this permission isn't required by anything else in this module.

When the data source isn't needed, then by the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege), we shouldn't require the `s3:ListAllMyBuckets` permission.

This additional permission is particularly obvious when [migrating](https://developer.hashicorp.com/terraform/tutorials/configuration-language/move-config) existing `aws_s3_*` resources into this module.

## Breaking Changes

Not a breaking change:

 - the module interface (i.e. input and output variables) has not changed; and,
 - for existing use cases where the IAM entity already has the `s3:ListAllMyBuckets` permission, the module will continue to behave as before except simply skip the `ListBuckets` S3 API calls.

## How Has This Been Tested?
- [x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s) - _No interface changes._

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,6 +1,8 @@
 data "aws_region" "current" {}
 
-data "aws_canonical_user_id" "this" {}
+data "aws_canonical_user_id" "this" {
+  count = local.create_bucket && ((var.acl != null && var.acl != "null") || length(local.grants) > 0) ? 1 : 0
+}
 
 data "aws_caller_identity" "current" {}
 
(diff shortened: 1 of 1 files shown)


theipster:
Bump please

antonbabenko:
This PR is included in version 3.15.0 :tada:

PR #261, opened 2023-11-21
feat: Allow override of the default tags override

kbcz1989:
## Description
Added support for ignoring default_tags set on provider level.

## Motivation and Context
This change was required as it was not possible to deploy s3 bucket object if there are more than 10 tags set on provider level.
More info [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_object#ignoring-provider-default_tags).

## Breaking Changes
This is not a breaking change but it does require bumping the provider version for 2 major versions, as option was only recently added to provider.

## How Has This Been Tested?


kbcz1989:
Let's try to keep it open for a bit more.

kbcz1989:
bump once more

modules/object/main.tf
```diff
@@ -36,4 +36,13 @@ resource "aws_s3_object" "this" {
   lifecycle {
     ignore_changes = [object_lock_retain_until_date]
   }
+
+  override_provider {
```
antonbabenko, line 40: Please change it so that the `override_provider` block is used only if necessary (`override_default_tags` should be `true`).

Also, please keep `lifecycle` block as the last one.

antonbabenko (changes requested):

antonbabenko:
This PR is included in version 4.1.0 :tada:

kbcz1989:
Hello @antonbabenko. Sorry for the delay and thank you for fixing and merging this!

PR #275, opened 2024-03-11
Draft: Establish a dedicated "NIST Mode" for easing compliance

apmarshall:
## Description
This PR introduces a new variable, `nist_mode`, which, when set to `true` overrides other variables within the module to guarantee certain values for compliance purposes.

In the context of this module, the changes made can be seen in the updates to the README, which documents both what requirements are being met and what variables are overridden to achieve this.

## Motivation and Context
There are some environments, especially those in the government space, where adherence to NIST SP 800-53 standards is required. Rather than creating separate modules that enforce these requirements, this allows users in those environments to continue using these community supported modules by setting a single variable to ensure compliance requirements are met.

## Breaking Changes
This should *not* introduce any breaking changes to uses who do not enable `nist_mode.` The default for `nist_mode` is set to false, which means that the module should continue to perform exactly as it currently does for users who have not enabled `nist_mode.` Enabling `nist_mode` overrides certain variables, as documented in the README, which *could* result in breaking changes for module implementations that would have been out of compliance before setting `nist_mode` to `true`.

## How Has This Been Tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,41 @@ module "s3_bucket_for_logs" {
 }
 ```
 
+## NIST Mode
+
+For government, FedRAMP, and other regulated environments who need to maintain compliance with NIST standards, enabling NIST Mode sets certain values within the module to achieve these requirements. In this module, setting `nist_mode = true` supports the following NIST SP-800-53r.5 requirements:
+- AC-2(4) -- Account Management: Automated Audit Actions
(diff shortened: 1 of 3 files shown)


antonbabenko:
Thank you for expressing interest. Closing for the same reasons as https://github.com/terraform-aws-modules/terraform-aws-kms/pull/29 .

PS: Shameless plug... https://compliance.tf is in development.

PR #299, opened 2024-12-20
feat: Added Account-level Public Access Block module, and minor fixes as fallback

antonbabenko:
* Added default value for versioning as fallback
* S3 account-level Public Access Block


antonbabenko:
This PR is included in version 4.3.0 :tada:

PR #302, opened 2025-01-12
feat: Add variable create_lambda_permission in notification module

### LINKED ISSUE #301: Add boolean varaible `aws_lambda_permission` to the Notifications module
## Is your request related to a new offering from AWS?
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - Yes ✅: a very long time

I want to define S3 Bucket Notifications that trigger Lambda functions that already have a Lambda Policy that allows the trigger. In this case I don't want this Notifications module to create the `aws_lambda_permission` resource.

Just like SNS and SQS Notifications have a Variable `create_{sqs|sns}_policy` there should be a boolean Variable `create_lambda_policy`

I don't think there's an alternative.

## Additional context
I will submit a Pull Request.

maxrabin:
## Description
In the Notification module create a new variable a new `create_lambda_policy` that functions similar to `create_sqs_policy` and `create_sns_policy`

## Motivation and Context

I define and manage my Lambda Policy separate from the Terraform where I manage my S3 Notifications. I don't want this module to force me to manage the Lambda policy here.

Fixes #301 

## Breaking Changes

I think that migrating `aws_lambda_permission.allow` from a single to a `count` this will cause the Resource to be recreated.
This can be avoided by performing [`terraform state mv`](https://developer.hashicorp.com/terraform/cli/commands/state/mv)

There are 3 types of notifications, and this change brings Lambda in line with SQS and SNS that support `create_sqs_policy` and `create_sns_policy`

## How Has This Been Tested?

### DIFF
--- a/modules/notification/README.md
+++ b/modules/notification/README.md
@@ -40,7 +40,7 @@ No modules.
 | <a name="input_bucket"></a> [bucket](#input\_bucket) | Name of S3 bucket to use | `string` | `""` | no |
 | <a name="input_bucket_arn"></a> [bucket\_arn](#input\_bucket\_arn) | ARN of S3 bucket to use in policies | `string` | `null` | no |
 | <a name="input_create"></a> [create](#input\_create) | Whether to create this resource or not? | `bool` | `true` | no |
-| <a name="input_create_lambda_policy"></a> [create\_lambda\_policy](#input\_create\_lambda\_policy) | Whether to create a policy for Lambda permissions or not? | `bool` | `true` | no |
+| <a name="input_create_lambda_permission"></a> [create\_lambda\_permission](#input\_create\_lambda\_permission) | Whether to create Lambda permissions or not? | `bool` | `true` | no |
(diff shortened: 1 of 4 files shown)


antonbabenko (changes requested):
Good idea, but let's rename var from `create_lambda_policy` to `create_lambda_permission` since Lambda has permissions, but SNS/SQS has policies.

antonbabenko:
This doesn't seem like a breaking change, since `for_each` is used, so we are good.

maxrabin:
> This doesn't seem like a breaking change, since `for_each` is used, so we are good.

I realized you're right!

maxrabin:
> Good idea, but let's rename var from `create_lambda_policy` to `create_lambda_permission` since Lambda has permissions, but SNS/SQS has policies.

I renamed it to `create_lambda_policy`

antonbabenko (approved):

antonbabenko:
This PR is included in version 4.4.0 :tada:

PR #303, opened 2025-01-20
feat: Add validation for lifecycle_rule along with terraform tests

okanmenevseoglu:
## Description
I have added a simple validation for **_lifecycle_rule_** that ensures correct configuration is passed to the module.

## Motivation and Context
In my company, we experienced that passing an incorrect map doesn't break the plan & apply pipelines but it also doesn't actually apply the required change. The main reason for this change is also that the configuration in this module is slightly different than what is explained on [aws_s3_bucket_lifecycle_configuration](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_lifecycle_configuration) which users usually confuse with.

## Breaking Changes
It is backwards compatible according to the tests I performed.

## How Has This Been Tested?
- [*] I have updated at least one of the `examples/*` to demonstrate and validate my change(s) 
**_There already are examples._**
**_I have also tested using terraform tests and a private project_**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,8 @@ repos:
     rev: v1.96.1
     hooks:
       - id: terraform_fmt
+        args:
+          - --args=-recursive
       - id: terraform_wrapper_module_for_each
       - id: terraform_docs
         args:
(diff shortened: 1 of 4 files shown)


antonbabenko:
This PR is just too much... Maybe it does the purpose, but it requires 99% refactoring... closing for now.

okanmenevseoglu:
@antonbabenko 
> This PR is just too much... Maybe it does the purpose, but it requires 99% refactoring... closing for now.
Could you please elaborate why this is too much and why it requires refactoring? I didn't get what you mean.

PR #307, opened 2025-01-23
feat: Add optional policy to deny ssec encrypted uploads

mbainter:
## Description
There have been incidents where leaked credentials have led to attackers encrypting S3 buckets with their own keys in a form of ransomware attack. AWS recommends implementing policy to prevent using SSEC if that is not a feature that you use in a given bucket. This PR adds an optional policy attachment that implements AWS' recommendation.

Reference:
https://aws.amazon.com/blogs/security/preventing-unintended-encryption-of-amazon-s3-objects/

## Motivation and Context
Enable users to easily block the use of SSEC in their buckets, mitigating a potential avenue of attack in the event of leaked credentials.

## How Has This Been Tested?
- [X] I have executed `pre-commit run -a` on my pull request

I have updated one of the examples, but I won't be able to test it until later.  I will try to follow up soon with the results.


antonbabenko (approved):
👍

mbainter:
As an FYI - I did just successfully test this, with and without the policy to make sure it applied correctly and that it does block attempts to put objects with SSEC.

PR #310, opened 2025-02-01
feat: Support S3 Directory Bucket

magreenbaum:
## Description
Support for directory buckets. 

## Motivation and Context
https://docs.aws.amazon.com/AmazonS3/latest/userguide/directory-buckets-overview.html
- https://github.com/hashicorp/terraform-provider-aws/issues/40626
- https://github.com/hashicorp/terraform-provider-aws/issues/34612

## Breaking Changes
No.

## How Has This Been Tested?


examples/directory-bucket/main.tf
```diff
@@ -0,0 +1,111 @@
+locals {
+  region  = "eu-west-1"
+  zone_id = "euw1-az1"
```
antonbabenko, line 3: Could you make AZ ID discoverable using a data source? There are cases where some AZs are not available in some accounts.

README.md
```diff
@@ -123,6 +123,7 @@ Users of Terragrunt can achieve similar results by using modules provided in the
 - [S3 Analytics](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/tree/master/examples/s3-analytics) - S3 bucket Analytics Configurations.
 - [S3 Inventory](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/tree/master/examples/s3-inventory) - S3 bucket Inventory configuration.
 - [S3 Account-level Public Access Block](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/tree/master/examples/account-public-access) - Manage S3 account-level Public Access Block.
+- [S3 Directory Bucket](https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/tree/master/examples/directory-bucket) - S3 Directory Bucket configuration.
```
antonbabenko, line 126: Please add "S3 Directory Bucket" to the features list at the beginning of this README also.

modules/directory-bucket/main.tf
```diff
@@ -0,0 +1,144 @@
+locals {
+  bucket_name = "${var.bucket_name_prefix}--${var.availability_zone_id}--x-s3"
+}
+
+resource "aws_s3_directory_bucket" "this" {
```
antonbabenko, line 5: This submodule creates one unique resource type (`aws_s3_directory_bucket`), and the rest are the same as those in the main S3 bucket module.

Given there are [rather big differences in features](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-express-differences.html), do you think we can have this resource `aws_s3_directory_bucket` in the main module and make the creation of it conditionally (e.g. `is_directory_bucket = true/false`)? This way, we will have less code repetition.

Or am I missing a bigger picture? What do you think?

antonbabenko (changes requested):

modules/directory-bucket/main.tf
magreenbaum: That works as well. I was initially thinking since most of the resources in the main module didn't apply to the directory bucket, it would be easier to have a sub-module, but using `is_directory_bucket` also works. I refactored to this. Let me know if that works.

antonbabenko (approved):

examples/directory-bucket/main.tf
```diff
@@ -0,0 +1,121 @@
...
+  bucket              = random_pet.this.id
+  availability_zone   = "${local.region}b"
+}
+
+module "complete" {
+  source = "../../"
+
+  is_directory_bucket = true
+  bucket              = "${random_pet.this.id}-complete"
+  availability_zone   = "${local.region}b"
```
antonbabenko, line 29: As I mentioned earlier. `b` or other AZs may not be present in some AWS accounts. I hit this issue in the past myself. Please update the example to use data-source to find first-available AZ in the account.

antonbabenko (changes requested):
Minor comment and we are good to go :) Thanks!

examples/directory-bucket/main.tf
magreenbaum: Oh I see, my apologies, I misunderstood initially. I added a data source to find available AZs but I found that even though an AZ is available to me, it might not be an option for directory buckets. In the example, if I set it to index `0`, it attempts to use `euw1-az2` which is available for my account but not available for directory buckets so I used index `1` here. https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-express-Endpoints.html

I guess this may not work for everyone so I'm happy to modify this.

antonbabenko (approved):

examples/directory-bucket/main.tf
antonbabenko: Thank you! 👍

antonbabenko:
This PR is included in version 4.6.0 :tada:

PR #312, opened 2025-02-10
feat: Allow for policy creation for externally created s3 buckets

### LINKED ISSUE #305: Circular dependency when trying to use `s3_bucket_id` when attaching a policy
I'm creating a Cloudfront distribution with OAC using the module `terraform-aws-modules/cloudfront/aws` and using the bucket name as the OAC config map ID.

I'm also creating the S3 bucket using this module to serve as the origin for the CloudFront distribution.

I'm also attaching an S3 bucket policy to the s3 Bucket by using the `attach_policy` and `policy` variables. The policy references both the S3 and the CF Distribution.

Since the policy resource is a separate resource from the s3_bucket resource there should be no circular dependency, but because the module output variable `s3_bucket_id` is linked to the policy first, I'm getting a circular dependency error in terraform.

Using a different string for the OAC config ID does not trigger this circular dependency error.

Using the bucket ID here is a requirement for us, since we have multiple origins in the CF distribution, and multiple CF distributions.

Here's some example code, extracted from our module that we use in multiple deployments:

```
data "aws_iam_policy_document" "contents_bucket_policy" {
  statement {
    sid    = "AllowCloudFrontServicePrincipal"
    effect = "Allow"
    principals {
      type        = "Service"
      identifiers = ["cloudfront.amazonaws.com"]
    }
    actions = [
      "s3:GetObject",
      "s3:ListBucket",
    ]
    resources = [
      module.contents_bucket.s3_bucket_arn,
      "${module.contents_bucket.s3_bucket_arn}/*"
    ]
    condition {
      test     = "StringEquals"
      variable = "AWS:SourceArn"
      values = [
        module.cloudfront.cloudfront_distribution_arn
      ]
    }
  }
}

module "contents_bucket" {
  source  = "terraform-aws-modules/s3-bucket/aws"
  version = "~> 4.0"

  attach_policy = true
  policy        = data.aws_iam_policy_document.contents_bucket_policy.json

# no other attach_* variables are being used here.

# [...]
}

module "cloudfront" {
  source  = "terraform-aws-modules/cloudfront/aws"
  version = "~> 4.0"

create_origin_access_control = true
  origin_access_control = {
    (module.contents_bucket.s3_bucket_id) = {
      description      = "CloudFront access to S3"
      origin_type      = "s3"
      signing_behavior = "always"
      signing_protocol = "sigv4"
    }
  }

# [...]

}
```

liame24:
## Description

Changes to the count which controls whether policy will be created/updated so that policies can be independently created for externally created s3 buckets, this is particularly useful if you use the module to create a bucket then create a cloudfront to serve the bucket and have to make adjustments to the policy after the bucket has been created as described in #305 

## Motivation and Context

Experienced the problem first hand and also resolves #305 

## Breaking Changes

This shouldn't break backwards compatibility

## How Has This Been Tested?

### DIFF
--- a/examples/s3-policy/README.md
+++ b/examples/s3-policy/README.md
@@ -1,0 +1,58 @@
+# Complete S3 bucket with most of supported features enabled
+
+This configuration allows for deployment of a S3 Bucket Policy independent of S3 Bucket creation
+
+## Usage
+
+To run this example you need to execute:
+
+```bash
+$ terraform init
+$ terraform plan
+$ terraform apply
+```
+
+Note that this example may create resources which cost money. Run `terraform destroy` when you don't need these resources.
+
+<!-- BEGIN_TF_DOCS -->
+## Requirements
+
+| Name | Version |
+|------|---------|
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.0 |
+| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 5.70 |
+| <a name="requirement_random"></a> [random](#requirement\_random) | >= 2.0 |
+
+## Providers
+
+| Name | Version |
+|------|---------|
+| <a name="provider_aws"></a> [aws](#provider\_aws) | >= 5.70 |
+| <a name="provider_random"></a> [random](#provider\_random) | >= 2.0 |
+
+## Modules
+
+| Name | Source | Version |
+|------|--------|---------|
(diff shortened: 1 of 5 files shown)


antonbabenko:
We will not support cases where the core resource (S3 bucket itself) is not created. It is a prerequisite for this module to do anything.

#305 should probably be solved another way; I don't have time to look into that one now.

PR #330, opened 2025-06-05
feat: Add `elb_log_delivery_policy_source_organizations` variable

szubersk:
## Description
* New `elb_log_delivery_policy_source_organizations` variable allows scoping access to the S3 bucket to AWS organization.

* Improve security by scoping ALB/NLB/VPC flow logs access to the caller's AWS account.

* Improve variable description. According to AWS documentation, the same S3 permissions can be used for both NLB access logs and VPC flow logs.

References:
* https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs-s3-permissions.html

* https://docs.aws.amazon.com/elasticloadbalancing/latest/network/enable-access-logs.html

* https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html

## Motivation and Context
* Reuse existing code to provide VPC flow logs delivery capability.
* Improve S3 access security.
* Improve documentation around attachable policies.

## Breaking Changes
N/A

## How Has This Been Tested?


main.tf
```diff
@@ -622,7 +622,7 @@ data "aws_iam_policy_document" "elb_log_delivery" {
     for_each = { for k, v in local.elb_service_accounts : k => v if k == data.aws_region.current.name }
 
     content {
-      sid = format("ELBRegion%s", title(statement.key))
```
bryantbiggs, line 625: this is a breaking change and should not be modified at this time

variables.tf
```diff
@@ -5,13 +5,13 @@ variable "create_bucket" {
 }
 
 variable "attach_elb_log_delivery_policy" {
-  description = "Controls if S3 bucket should have ELB log delivery policy attached"
```
bryantbiggs, line 8: why are there all of these changes? ELB is correct - ALB and NLB are in fact ELB v2 (classic load balancer was ELB v1)

bryantbiggs (changes requested):
it looks like we are trying to do too much in this PR and I'm not quite following, nor do I agree with the name changes (i.e. `ELB` to `ALB`, etc.)

PR #355, opened 2025-08-21
feat: Add tas_all in output.tf

paikend:
# Pull Request Description

## Description
Added `s3_bucket_tags_all` output to enable validation of S3 bucket tags in Terraform tests. This output exposes the complete tag set applied to S3 buckets for testing purposes.

## Motivation and Context
The existing outputs.tf file was missing the `tags_all` output, which is required for tag validation in Terraform tests. This change enables automated validation of required tags as part of the CI/CD pipeline.

## Breaking Changes
This is a non-breaking change as it only adds a new output without modifying existing functionality.

## How Has This Been Tested?
- [x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects
- [x] I have executed `pre-commit run -a` on my pull request

Added a simple test case to validate that the Environment tag exists on S3 buckets:

```hcl
run "validate_s3_required_tags" {
  command = plan
  
  assert {
    condition     = alltrue([for bucket in module.s3_bucket : contains(keys(bucket.s3_bucket_tags_all), "Environment")])
    error_message = "S3 Bucket is missing required 'Environment' tag."
  }
}
```

This test ensures that all S3 buckets have the required Environment tag applied.

PR #357, opened 2025-08-27
feat: Add tags in output.tf #355

paikend:
## Description
Added `s3_bucket_tags` output to expose user-defined tags applied to S3 buckets. This output provides access to the tags that users explicitly set via the `tags` variable, enabling tag validation and testing in downstream configurations.

## Motivation and Context
The module was missing an output for S3 bucket tags, which is commonly needed for:
- Automated tag validation in Terraform tests
- Tag consistency checks across infrastructure
- Integration with other modules that need to reference bucket tags
- Compliance and governance validation

This addresses the need for accessing user-defined bucket tags without exposing computed-only values like `tags_all`.

## Breaking Changes
This is a **non-breaking change**. It only adds a new output without modifying any existing functionality or outputs.

## How Has This Been Tested?

### Testing Details
- Validated the output returns the expected tags in the complete example
- Confirmed the output works correctly when no tags are provided (returns empty object)
- Tested integration with conditional bucket creation scenarios using the `try()` function
- Added test case to validate tag accessibility:

```hcl
run "validate_tags_output" {
  command = plan
  
  variables {
    tags = {
      Environment = "test"
      Project     = "terraform-aws-s3-bucket"
    }
  }
  
  assert {
    condition = module.s3_bucket.s3_bucket_tags.Environment == "test"
    error_message = "S3 bucket tags output should include Environment tag"
  }
  
  assert {
    condition = length(keys(module.s3_bucket.s3_bucket_tags)) == 2
    error_message = "S3 bucket tags output should contain exactly 2 tags"
  }
}
```

The output safely handles edge cases using `try(aws_s3_bucket.this[0].tags, {})` to return an empty object when the bucket resource doesn't exist.


antonbabenko:
Looks pretty good, please fix CI before the merge.

paikend:
Sorry, I Fixed it

PR #375, opened 2026-01-25
fix(lifecycle): filter null values from dynamic blocks

rorynolan:
## Description

When lifecycle_rule attributes (`expiration`, `transition`, `noncurrent_version_expiration`, `noncurrent_version_transition`, `filter`) are `null`, `flatten([null])` returns `[null]` (not `[]`), causing dynamic blocks to iterate once with a null value.

### Problems this causes:

1. **expiration / noncurrent_version_expiration**: Creates empty blocks with default values (`days=0`, `expired_object_delete_marker=false`). AWS rejects `expired_object_delete_marker` when combined with object size filters, causing apply failures even when expiration wasn't specified.

2. **transition / noncurrent_version_transition**: Creates empty blocks that could cause unexpected behavior or API errors when `storage_class` is required but missing.

3. **filter**: When filter is `null`, the empty filter block isn't created (since `length([null]) == 1`, not `0`), requiring users to work around it by passing `filter = {}` instead of omitting it.

### Fix

Filter null values from flattened lists before iteration:

```hcl
# Before
for_each = try(flatten([rule.value.expiration]), [])

# After  
for_each = [for v in try(flatten([rule.value.expiration]), []) : v if v != null]
```

This preserves backwards compatibility for list inputs while correctly handling null values.

### Files changed

- `main.tf`: Updated 8 dynamic block `for_each` expressions:
  - `expiration`
  - `transition`
  - `noncurrent_version_expiration`
  - `noncurrent_version_transition`
  - 3 `filter` blocks in lifecycle configuration
  - 3 `filter` blocks in replication configuration

### Testing

Tested with a real Terraform configuration that previously failed with:
```
Error: api error InvalidRequest: ExpiredObjectDeleteMarker cannot be specified with Object Size.
```

After this fix, the configuration applies successfully without requiring any workarounds in the calling module.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -321,7 +321,7 @@ resource "aws_s3_bucket_lifecycle_configuration" "this" {
 
       # Max 1 block - expiration
       dynamic "expiration" {
-        for_each = try(flatten([rule.value.expiration]), [])
+        for_each = [for v in try(flatten([rule.value.expiration]), []) : v if v != null]
 
         content {
           date                         = try(expiration.value.date, null)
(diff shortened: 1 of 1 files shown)


bryantbiggs:
start with an issue and a reproduction before we try to turn Claude loose

rorynolan:
Fair enough. I've done this now in #376

PR #380, opened 2026-02-24
fix: Removed object_lock_enabled variable in aws_s3_bucket resource

nronix:
## Description
The module should follow what aws console offers. aws_s3_bucket_object_lock_configuration will handle object lock enable and disable, default retention etc. Using this variable here was causing resources to be forced replaced / errors in pipeline with newer versions of terraform

## Motivation and Context

The Change will prevent deletion of s3 bucket in case object lock value is update i.e either from true to false or false to true.aws_s3_bucket_object_lock_configuration works as expected and matches the AWS console. 
Also this is deprecated as per Terraform documentation https://registry.terraform.io/providers/-/aws/6.33.0/docs/resources/s3_bucket

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -51,7 +51,11 @@ resource "aws_s3_bucket" "this" {
   bucket_prefix = var.bucket_prefix
 
   force_destroy       = var.force_destroy
-  object_lock_enabled = var.object_lock_enabled
+  /* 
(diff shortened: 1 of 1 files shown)


bryantbiggs:
https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/issues/381#issuecomment-3953040630

PR #385, opened 2026-04-04
feat: Support null value for replication_configuration (Closes #379)

### LINKED ISSUE #379: Enable support for conditional replication (declarative)
I'm working on a module to deploy an S3 bucket with various settings.
The module needs to accept a parameter that indicates whether replication should be enabled.

`variables.tf`
```
variable "enable_replication" {
  type = bool
  default = false
}
```

`main.tf`
``` tf
module "bucket" {
  source = "..."

  replication_configuration = var.enable_replication ? {
      role = "..."
      rules = [...]
  } : null
}
```

This is not supported as the `replication_configuration` object is not expected to be `null`.

Idk if this would break other parts of the framework, but could be fixed by changing this condition:
```
resource "aws_s3_bucket_replication_configuration" "this" {
  count = local.create_bucket && length(keys(var.replication_configuration)) > 0 && !var.is_directory_bucket ? 1 : 0
}

resource "aws_s3_bucket_replication_configuration" "this" {
  count = local.create_bucket && var.replication_configuration != null && length(keys(var.replication_configuration)) > 0 && !var.is_directory_bucket ? 1 : 0
}
```

hasansyedCS:
## Description
Support `null` value for `replication_configuration` variable to allow clean conditional enabling of replication.

## Motivation and Context
Closes #379

Many users want to conditionally enable replication using a boolean flag, like this:

`replication_configuration = var.enable_replication ? {`
`role = "..."`
`rules = [...]`
`} : null`

Previously, passing null caused an error because length(keys(null)) is invalid in the count meta-argument.
This change makes the common conditional pattern work while keeping all existing behavior unchanged.

## Breaking Changes
None. This is fully backwards compatible.

## How Has This Been Tested?

 I have tested and validated these changes using one or more of the provided examples/* projects
 I have executed pre-commit run -a on my pull request (if applicable)

## Testing done:

Ran terraform validate successfully
Confirmed existing behavior with replication_configuration = {} still works
Verified the new conditional null pattern now works without error

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -435,7 +435,7 @@ resource "aws_s3_bucket_object_lock_configuration" "this" {
 }
 
 resource "aws_s3_bucket_replication_configuration" "this" {
-  count = local.create_bucket && length(keys(var.replication_configuration)) > 0 && !var.is_directory_bucket ? 1 : 0
+  count = local.create_bucket && var.replication_configuration != null && length(keys(var.replication_configuration)) > 0 && !var.is_directory_bucket ? 1 : 0
 
   region = var.region
 
--- a/variables.tf
+++ b/variables.tf
@@ -239,7 +239,7 @@ variable "lifecycle_rule" {
 }
 
 variable "replication_configuration" {
-  description = "Map containing cross-region replication configuration."
+  description = "Map containing cross-region replication configuration. Set to `null` to disable replication (useful for conditional enabling)."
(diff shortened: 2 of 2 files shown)


bryantbiggs:
https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/issues/379#issuecomment-4235982006

PR #388, opened 2026-05-01
feat: Add Amazon S3 Files submodule

### LINKED ISSUE #386: Proposal: add an Amazon S3 Files submodule when provider support is available
### Summary

AWS recently announced Amazon S3 Files, which provides file-system access to data stored in an existing S3 bucket.

From the AWS documentation and the Terraform AWS provider tracking issue, this appears to be a separate service/resource area rather than a simple bucket setting:
- https://aws.amazon.com/about-aws/whats-new/2026/04/amazon-s3-files/
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-files.html
- https://github.com/hashicorp/terraform-provider-aws/issues/47324

### Proposal

Once Terraform AWS provider support is available, would you consider adding an `s3-files` submodule under this repository rather than extending the root `terraform-aws-s3-bucket` module directly?

### Rationale

`terraform-aws-s3-bucket` already has related but separate submodules under `modules/`, so this seems like a natural fit for the existing repository structure.

It also seems like a cleaner separation of concerns:

- the root module can remain focused on bucket-native configuration
- an `s3-files` submodule could manage S3 Files-specific resources and wiring

From the AWS docs, S3 Files seems to introduce a separate resource graph around:
- a file system linked to an S3 bucket or prefix
- IAM roles/policies
- mount targets / networking
- compute-side integration

Because of that, this feels adjacent to the bucket module, but not necessarily a good fit inside the root module itself.

### Possible Direction

A composition model could look something like:

- root `terraform-aws-s3-bucket` module
  - bucket creation and bucket-native configuration
  - prerequisites such as versioning and encryption

- `modules/s3-files`
  - S3 Files file system resources
  - related IAM
  - mount targets / networking as appropriate
  - outputs that can be consumed by EC2/EKS-style examples

Given the existing submodules in this repository, an `s3-files` submodule seems like the most natural fit to me.

If maintainers feel this is too separate from the rest of the repository, a standalone `terraform-aws-s3-files` module could also make sense.

### Additional Context

Existing submodules in this repository make this pattern seem plausible:
- https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/tree/master/modules

If this direction seems useful, I'd be happy to help create the submodule once the provider resources are available.

sharansutrapu:
## Description
This PR adds the `s3-files` submodule mapping to the newly released `aws_s3files_*` provider resources. 

Following the repository's established submodule patterns, this implementation:
- Is strictly decoupled from the root bucket module.
- Supports multi-AZ deployments by dynamically creating mount targets across provided subnets.
- Includes strict `lifecycle` preconditions for safety (validating ARNs, VPC/Subnet mappings, and avoiding duplicate subnets).
- Falls back to a secure, VPC-scoped default `aws_s3files_file_system_policy` if a custom JSON policy is not provided.
- Explicitly maps `depends_on = [module.s3_bucket]` within the example deployment to prevent the `BucketHasS3FileSystemAttached` versioning race condition during `terraform destroy`.
- Includes comprehensive `README.md` files for both the module and the example.

## Motivation and Context
Closes #386. AWS recently released Amazon S3 Files, and this adds the corresponding isolated submodule to manage those resources alongside existing buckets.

## Breaking Changes
None. This is a purely additive submodule that does not modify the root bucket logic.

## How Has This Been Tested?
- [ ] I have executed `pre-commit run -a` on my pull request (Left for CI)

I have successfully initialized, planned, applied, and destroyed the `examples/s3-files` directory against a live AWS environment (`eu-west-1`). The submodule correctly maps dependencies to the root bucket, provisions the file system and mount targets, and tears them down cleanly without locking.

<details>
<summary>Click to view terraform apply output</summary>

```text
Apply complete! Resources: 14 added, 0 changed, 0 destroyed.

Outputs:

s3_bucket_arn = "arn:aws:s3:::s3-files-vital-oarfish"
s3_bucket_id = "s3-files-vital-oarfish"
s3_bucket_versioning_status = "Enabled"
s3_files_file_system_arn = "arn:aws:s3files:eu-west-1:600627358256:file-system/fs-0ad5c5ddf95f324d6"
s3_files_file_system_id = "fs-0ad5c5ddf95f324d6"
s3_files_mount_target_network_interface_ids = [
  "eni-04fae257e1070d958",
  "eni-0206d68755fbee291",
]
```
</details>

<details>
<summary>Click to view terraform destroy output</summary>

```text
module.s3_bucket.aws_s3_bucket.this[0]: Destruction complete after 0s
random_pet.this: Destroying... [id=vital-oarfish]
random_pet.this: Destruction complete after 0s
aws_iam_role.s3_files: Destruction complete after 1s
aws_security_group.s3_files: Destruction complete after 2s
aws_vpc.this: Destroying... [id=vpc-0c22b91203fce495a]
aws_vpc.this: Destruction complete after 0s

Destroy complete! Resources: 14 destroyed.
```
</details>


milosveljkovic (commented):
why there is no aws_s3files_access_point ?

sharansutrapu:
Great catch, @milosveljkovic ! I started with the core filesystem and mount target setup to get the foundational wiring in place, but you’re absolutely right, Access Points are an important part of enforcing least privilege POSIX identities properly.

I’ll add an "access_points" map variable along with the "aws_s3files_access_point" resource and push an updated version shortly.

sharansutrapu:
Great catch, @milosveljkovic! I completely agree that Access Points are crucial for enforcing least-privilege POSIX identities.

I've just pushed a comprehensive update that addresses this and adds a few proactive Day-2 operational fixes:

1. Access Points (feat): Added the aws_s3files_access_point resource mapped to an access_points map variable using a for_each loop. Updated the examples/s3-files deployment to demonstrate creating one with POSIX user/group controls.
2. Day 2 JSON Drift Prevention (fix): Wrapped the custom file_system_policy input in a jsonencode(jsondecode(...)) round-trip. AWS normalizes raw JSON strings to compact form on store, causing a perpetual diff on subsequent terraform plan runs. This normalizes the user's input safely before it hits state.
3. Example NFS Connectivity (fix): Added the missing TCP port 2049 (NFS) ingress rule to the example's security group so the mount targets are reachable.
4. Output Completeness (fix): Ensured the s3_files_access_points output returns the name attribute, as clients will need it to construct their mount paths.
5. Docs Sync: Updated both README.md files to document the new access_points variable, the NFS ingress rule, and the provider version bump to 6.45.0.

Everything was re-tested (plan, apply, destroy) against a live AWS environment. Let me know if you need any other tweaks!

modules/s3-files/versions.tf
```diff
@@ -0,0 +1,16 @@
+terraform {
+  required_version = ">= 1.5.7"
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 6.39"
```
milosveljkovic: lets bump this to '>= 6.42'

milosveljkovic: align with other modules

modules/s3-files/variables.tf
```diff
@@ -0,0 +1,101 @@
...
+  default     = true
+}
+
+variable "region" {
+  description = "Region where the resource(s) will be managed. Defaults to the region set in the provider configuration"
+  type        = string
+  default     = null
+}
+
+variable "s3_uri" {
```
milosveljkovic: I would rename this to bucket_arn

milosveljkovic: s3_uri is confusing  b/c it is usually referenced to 's3://name-of-bucket'

modules/s3-files/main.tf
```diff
@@ -0,0 +1,284 @@
...
+  lifecycle {
+    # Guard 1 – s3_uri must be a well-formed S3 bucket ARN.
+    # S3 ARNs take the form arn:<partition>:s3:::<bucket-name> with no region
+    # or account-id segments.  Bucket names are 3–63 characters and may only
+    # contain lowercase letters, numbers, hyphens, and dots.
+    precondition {
+      condition = (
+        var.s3_uri != null &&
+        trimspace(var.s3_uri) != "" &&
+        can(regex("^arn:[^:]+:s3:::[A-Za-z0-9][A-Za-z0-9.-]{1,61}[A-Za-z0-9]$", var.s3_uri))
```
milosveljkovic: This should support new bucket type with 'Account Regional namespace'

sharansutrapu:
Hi @milosveljkovic, thanks for the review!

I've applied all your suggestions:

Bumped the AWS provider minimum version to >= 6.42 to align with the rest of the repository.

Renamed s3_uri to bucket_arn across the submodule, examples, and docs for clarity.

Removed the 63-character limit from the ARN validation regex so it safely accommodates the longer names generated by the new Account Regional namespace buckets.

Let me know if you need anything else to get this over the finish line!

sharansutrapu:
Hi @milosveljkovic ,

Just giving this a quick bump! I have rebased the branch against the latest master to keep everything perfectly up to date. I also ran the pre-commit hooks to resolve a minor terraform-docs formatting drift that was failing the CI pipeline.

antonbabenko:
@magreenbaum you know this module and our module standards very well, so if you have time, could you please take a look at this and finish this PR?

modules/s3-files/main.tf
```diff
@@ -0,0 +1,284 @@
...
+  lifecycle {
+    # Guard 1 – bucket_arn must be a well-formed S3 bucket ARN.
+    # S3 ARNs take the form arn:<partition>:s3:::<bucket-name> with no region
+    # or account-id segments.  Standard bucket names are 3–63 characters;
+    # Account Regional buckets may be up to 255 characters.
+    precondition {
+      condition = (
+        var.bucket_arn != null &&
+        trimspace(var.bucket_arn) != "" &&
+        can(regex("^arn:[^:]+:s3:::[A-Za-z0-9][A-Za-z0-9.-]+[A-Za-z0-9]$", var.bucket_arn))
```
milosveljkovic: not sure if we need upper cases here since those not allowed in s3 names

milosveljkovic (commented):
Are we missing here the aws_s3files_synchronization_configuration? It would be nice to have it here as well with the following defaults:

```tf
variable "synchronization_import_data_rules" {
  description = "Import data rules for S3 Files synchronization."
  type = list(object({
    prefix         = string
    size_less_than = number
    trigger        = string
  }))

  default = [
    {
      prefix         = ""
      size_less_than = 131072 // 128KB - can we set this as 128KB or 10MB instead of bytest (will be much easier for end-user I guess)
      trigger        = "ON_FILE_ACCESS"
    }
  ]
}

variable "synchronization_expiration_data_rule" {
  description = "Expiration rule for S3 Files synchronization."
  type = object({
    days_after_last_access = number
  })
  // got this from aws as they set that as default
  default = 30
}
```
etc...

modules/s3-files/variables.tf
```diff
@@ -0,0 +1,101 @@
+variable "create" {
+  description = "Whether to create s3 files resources"
```
milosveljkovic: nit: Whether to create this resource or not?

lets use this syntact to align with other modules if possible, not a deal breaker to leave it as is

modules/s3-files/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+Because the S3 Files file system holds a reference to the S3 bucket, attempting to destroy the bucket before the file system is detached results in a `409 BucketHasS3FileSystemAttached` error.  When managing both resources in the same Terraform configuration, add `module.s3_bucket` to `depends_on` on the `module "s3_files"` block:
+
+```hcl
+module "s3_files" {
+  source = "terraform-aws-modules/s3-bucket/aws//modules/s3-files"
+
+  # ...
+
+  depends_on = [
```
milosveljkovic, line 35: is it ok this to be default behaviour (add it to the aws_s3files_file_system resource?)  or at least to put it in the example

sharansutrapu:
Thanks for the review, @milosveljkovic ! I've applied all of the feedback:

Regex: Updated the ARN validation to strictly enforce lowercase [a-z0-9].

Descriptions: Updated the create description to match the repository standard.

Sync Config: Added the aws_s3files_synchronization_configuration resource with the suggested defaults. (I kept size_less_than as a number since the provider natively expects bytes, but added a note in the description that 131072 translates to 128KB for clarity).

Depends On: Regarding your question about the depends_on = [module.s3_bucket], to avoid tightly coupling the submodule to the root module, I didn't hardcode it in the submodule itself. However, I did explicitly include it in the examples/s3-files/main.tf deployment so users have a safe, working reference to avoid that race condition.

Everything is pushed up and CI is green. Let me know if we are good to go!

milosveljkovic (approved):
tyvm, seems ok

modules/s3-files/main.tf
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    # Guard 9 – IPv6 address map keys must be a subset of subnet_ids.
+    precondition {
+      condition = length(
+        setsubtract(toset(keys(var.mount_target_ipv6_addresses)), toset(var.subnet_ids))
+      ) == 0
+      error_message = "mount_target_ipv6_addresses keys must match values provided in subnet_ids."
+    }
+  }
```
magreenbaum: Lets leave out these lifecycle precondition blocks and their associated data sources. This adds additional API calls and more maintenance overhead for the module. Generally not standard practice in these modules either.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,305 @@
...
+  region = var.region
+
+  file_system_id = aws_s3files_file_system.this[0].id
+  # jsonencode(jsondecode(...)) normalises any caller-supplied JSON to the same
+  # compact canonical form that the AWS API returns on reads.  Without this
+  # round-trip, pretty-printed or differently-ordered policy strings cause a
+  # perpetual diff on every subsequent plan even though the policy has not
+  # changed.  data.aws_iam_policy_document already emits canonical JSON so the
+  # auto-generated path needs no normalisation.
+  policy = var.file_system_policy != null ? jsonencode(jsondecode(var.file_system_policy)) : data.aws_iam_policy_document.this[0].json
```
magreenbaum: These jsonencode then jsondecode functions together here seems strange. Is this an issue with this resource argument specifically? If not, lets pass it through as-is and let the user handle inputting valid json:
```suggestion
  policy = var.file_system_policy != null ? var.file_system_policy : data.aws_iam_policy_document.this[0].json
```

modules/s3-files/main.tf
```diff
@@ -0,0 +1,305 @@
...
+
+  lifecycle {
+    # Validate that a custom policy string is parseable JSON before applying.
+    # jsondecode() throws on invalid JSON; can() converts that to a bool.
+    precondition {
+      condition     = var.file_system_policy == null || can(jsondecode(var.file_system_policy))
+      error_message = "file_system_policy must be valid JSON when provided."
+    }
+  }
```
magreenbaum: Lets remove this lifecycle block and let the user handle validating the json that their passing in.

modules/s3-files/variables.tf
```diff
@@ -0,0 +1,127 @@
...
+
+variable "tags" {
+  description = "Key-value map of resource tags"
+  type        = map(string)
+  default     = {}
+}
+
+variable "access_points" {
+  description = "Map of S3 Files access point configurations to create"
+  type        = any
```
magreenbaum: Lets use specific type expressions for this instead of `any`.

examples/s3-files/main.tf
```diff
@@ -0,0 +1,197 @@
...
+resource "aws_subnet" "private_b" {
+  vpc_id                  = aws_vpc.this.id
+  cidr_block              = "10.42.2.0/24"
+  availability_zone       = data.aws_availability_zones.available.names[1]
+  map_public_ip_on_launch = false
+
+  tags = merge(local.tags, {
+    Name = "s3-files-example-private-b"
+  })
+}
```
magreenbaum: You could use the vpc module here in place of these standalone resources: https://github.com/terraform-aws-modules/terraform-aws-vpc

examples/s3-files/versions.tf
```diff
@@ -0,0 +1,14 @@
+terraform {
+  required_version = ">= 1.5.7"
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 6.39"
```
magreenbaum: We should match the example aws provider version with the module it uses and rest of the repo. 
```suggestion
      version = ">= 6.42"
```

modules/s3-files/main.tf
```diff
@@ -0,0 +1,305 @@
...
+# All fields except file_system_id are optional; dynamic blocks are skipped
+# when the corresponding key is absent from the entry map.
+resource "aws_s3files_access_point" "this" {
+  for_each = var.create ? var.access_points : {}
+
+  region = var.region
+
+  file_system_id = aws_s3files_file_system.this[0].id
+
+  tags = try(merge(var.tags, each.value.tags), var.tags)
```
magreenbaum: Once we use specific type expressions for `access_points`, we can drop the `try` function for this.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,305 @@
...
+
+  # posix_user is required by the provider schema - it is a static block, not
+  # dynamic.  Using a static block here makes the requirement visible in HCL
+  # and avoids the null-access path that a dynamic block with try() introduces
+  # when the caller explicitly passes posix_user = null.
+  posix_user {
+    uid            = each.value.posix_user.uid
+    gid            = each.value.posix_user.gid
+    secondary_gids = try(each.value.posix_user.secondary_gids, null)
```
magreenbaum: Same for here and the others below in this same resource, we can drop the `try` functions once we use specific type expressions for `access_points`.

sharansutrapu:
Thanks for the detailed review, @magreenbaum! I appreciate the guidance on the module standards here. 

I have pushed an update that implements all of your feedback:
- Removed the `lifecycle` precondition blocks and associated data sources to reduce API overhead.
- Removed the `jsonencode/decode` wrapper and JSON validation from the policy.
- Implemented strict `map(object({ ... }))` typing with `optional()` fields for the `access_points` variable.
- Stripped the `try()` functions from the access point resource now that strict typing is enforced.
- Updated the example deployment to use the `terraform-aws-modules/vpc/aws` module and bumped the example provider version to `>= 6.42`.
- Ran `pre-commit` locally to ensure all formatting and `terraform-docs` injection is up to date.

Let me know if everything looks good to go!

modules/s3-files/variables.tf
```diff
@@ -0,0 +1,142 @@
...
+  type = list(object({
+    prefix         = string
+    size_less_than = number
+    trigger        = string
+  }))
+  default = [
+    {
+      prefix         = ""
+      size_less_than = 131072
+      trigger        = "ON_FILE_ACCESS"
```
magreenbaum: Any reason we're going with this as the default? Looks like `ON_DIRECTORY_FIRST_ACCESS` is the default aws will set: 
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-files-synchronization-customizing.html
- https://github.com/hashicorp/terraform-provider-aws/blob/706bf4005c9831858dae6dbebd9476128032f0fe/internal/service/s3files/synchronization_configuration.go#L234

modules/s3-files/main.tf
```diff
@@ -0,0 +1,188 @@
...
+
+  region = var.region
+
+  file_system_id = aws_s3files_file_system.this[0].id
+  # jsonencode(jsondecode(...)) normalises any caller-supplied JSON to the same
+  # compact canonical form that the AWS API returns on reads.  Without this
+  # round-trip, pretty-printed or differently-ordered policy strings cause a
+  # perpetual diff on every subsequent plan even though the policy has not
+  # changed.  data.aws_iam_policy_document already emits canonical JSON so the
+  # auto-generated path needs no normalisation.
```
magreenbaum: Lets remove this comment also since we removed the jsonencode/decode functions.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,188 @@
...
+        }
+      }
+    }
+  }
+}
+
+# Attaches either the caller-supplied policy JSON or the auto-generated
+# secure-default policy to the file system.  The precondition rejects
+# caller-supplied JSON that cannot be parsed, producing a clear error at plan
+# time rather than a confusing AWS API error at apply time.
```
magreenbaum: Also here, we should remove or update this comment since we removed the precondition.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,188 @@
...
+    for_each = each.value.root_directory != null ? [each.value.root_directory] : []
+
+    content {
+      path = root_directory.value.path
+
+      dynamic "creation_permissions" {
+        # Filter explicit nulls: try() only catches missing keys; if the caller
+        # writes creation_permissions = null the list would be [null] and the
+        # content block would crash on .owner_uid.  The `if v != null` guard
+        # collapses that to an empty list, skipping the block entirely.
```
magreenbaum: Lets remove this comment as well since we're not using try function here anymore.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,188 @@
...
+  region = var.region
+
+  file_system_id = aws_s3files_file_system.this[0].id
+
+  tags = merge(var.tags, each.value.tags)
+
+  # posix_user is required by the provider schema - it is a static block, not
+  # dynamic.  Using a static block here makes the requirement visible in HCL
+  # and avoids the null-access path that a dynamic block with try() introduces
+  # when the caller explicitly passes posix_user = null.
```
magreenbaum: Update or remove comment since we're not using try function anymore.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,188 @@
+# Aliases subnet_ids into a local to allow future enrichment (e.g. sorting or
+# filtering) without touching the resource count expressions below.
+locals {
+  subnet_ids = var.subnet_ids
+}
+
+# The central S3 Files file system resource.  A single file system is created
+# per invocation of this module; mount targets (one per subnet) are managed
+# separately below.  The lifecycle preconditions run at plan-time and catch
+# mis-configurations before any AWS API call is made.
```
magreenbaum: Lets remove or update this since lifecycle preconditions were removed.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,188 @@
...
+      ]
+    }
+
+    resources = ["*"]
+
+    # The aws:SourceVpc condition is the primary network-layer guard: it prevents
+    # mount requests that do not originate from inside the declared VPC, even if
+    # an identity policy would otherwise permit them.  The block is conditional so
+    # that the policy remains valid when vpc_id is somehow null (unlikely given
+    # the precondition above, but defensive programming is good practice).
```
magreenbaum: Remove or update comment to remove info about the precondition.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,188 @@
+# Aliases subnet_ids into a local to allow future enrichment (e.g. sorting or
+# filtering) without touching the resource count expressions below.
```
magreenbaum: I'm not sure a lot of these comments in here are necessary and will be required to be maintained. For me I wouldn't add comments unless there's a quirk about the module's functionality it is describing. But no strong objection as long as they're accurate.

magreenbaum (commented):
@sharansutrapu thanks for the updates! A couple more small items. 

Lets also update the main README feature list and usage section: https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/blob/master/README.md

sharansutrapu:
Thanks @magreenbaum! I've pushed an update to address these final cleanup items:

- Updated the root `README.md` to include the S3 Files submodule in the feature list and usage section.
- Changed the `synchronization_import_data_rules` default trigger to `ON_DIRECTORY_FIRST_ACCESS` to perfectly align with AWS defaults.
- Scrubbed all the stale comments in `main.tf` related to the previously removed `jsonencode/decode`, `try()` functions, and preconditions to keep the codebase clean and minimalist.
- Ran the `pre-commit` hooks locally to ensure all docs and wrappers are perfectly synced.

Everything looks green on my end. Let me know if we're good to merge!

magreenbaum (approved):
@antonbabenko PR looks good.

sharansutrapu:
Thanks for all your time and the thorough reviews, @magreenbaum

examples/s3-files/README.md
```diff
@@ -0,0 +1,90 @@
+# S3 Files Example
+
+Configuration in this directory demonstrates a complete, production-ready deployment of the [`modules/s3-files`](../../modules/s3-files) submodule against a real AWS account.
+
+The example was validated against AWS provider `6.45.0` in `eu-west-1`.  All resources were successfully applied and cleanly destroyed (see [apply.md](apply.md) and [destroy.md](destroy.md) for the captured CLI output).
```
antonbabenko:
```suggestion
```

antonbabenko: Useless info.

examples/s3-files/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+## What this example creates
```
antonbabenko: Drop this whole section and table - there is already terraform-docs sections below with the same info

modules/s3-files/main.tf
```diff
@@ -0,0 +1,169 @@
+
+locals {
+  subnet_ids = var.subnet_ids
```
antonbabenko: Useless assignment, please drop locals block and empty line at the beginning.

examples/s3-files/main.tf
```diff
@@ -0,0 +1,174 @@
...
+
+resource "aws_iam_role" "s3_files" {
+  name_prefix        = "s3-files-example-"
+  assume_role_policy = data.aws_iam_policy_document.s3_files_assume_role.json
+
+  tags = local.tags
+}
+
+module "s3_bucket" {
```
antonbabenko: Move the main module calls (s3-bucket and s3-files) above the extra resources/modules. See other modules examples.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,169 @@
...
+}
+
+# The central S3 Files file system resource.  A single file system is created
```
antonbabenko: Useless comment. Please remove it.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,169 @@
...
+  kms_key_id            = var.kms_key_id
+  prefix                = var.prefix
+
+  tags = var.tags
+}
+
+# One mount target is created per subnet entry, enabling multi-AZ deployments
+# from a single module invocation.  Using count with an index into local.subnet_ids
+# (rather than for_each) keeps the resource address stable when subnets are
+# reordered but preserves Terraform's predictable destroy-before-create ordering.
```
antonbabenko: Lets use `for_each`, since its main point - allow to change list items without recreating the rest of resources. Also, please do not write such verbose comments at all! See other modules for examples!

modules/s3-files/main.tf
```diff
@@ -0,0 +1,169 @@
...
+  ip_address_type = var.ip_address_type
+  ipv4_address    = try(var.mount_target_ipv4_addresses[local.subnet_ids[count.index]], null)
+  ipv6_address    = try(var.mount_target_ipv6_addresses[local.subnet_ids[count.index]], null)
+  security_groups = var.security_group_ids
+}
+
+# ---------------------------------------------------------------------------
+# Default file system policy (auto-generated when file_system_policy is null)
+# ---------------------------------------------------------------------------
+# These three data sources are only evaluated when a custom policy is NOT
```
antonbabenko: Same with comments here - do not explain what this do and why. It is immediately obvious.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,169 @@
...
+    principals {
+      type = "AWS"
+      identifiers = [
+        "arn:${data.aws_partition.this[0].partition}:iam::${data.aws_caller_identity.this[0].account_id}:root"
+      ]
+    }
+
+    resources = ["*"]
+
+    # The aws:SourceVpc condition is the primary network-layer guard: it prevents
```
antonbabenko: yes, drop these noisy comments also, please. No need to repeat facts from the AWS security here.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,169 @@
...
+      content {
+        test     = "StringEquals"
+        variable = "aws:SourceVpc"
+        values   = [condition.value]
+      }
+    }
+  }
+}
+
+# One access point per entry in var.access_points.  Access points allow
```
antonbabenko: Drop comments, please

modules/s3-files/main.tf
```diff
@@ -0,0 +1,169 @@
...
+          owner_uid   = creation_permissions.value.owner_uid
+          owner_gid   = creation_permissions.value.owner_gid
+          permissions = creation_permissions.value.permissions
+        }
+      }
+    }
+  }
+}
+
+# Attaches either the caller-supplied policy JSON or the auto-generated
```
antonbabenko: Drop the comments, please

modules/s3-files/README.md
```diff
@@ -0,0 +1,238 @@
...
+
+### Basic
+
+Minimum viable configuration — creates one mount target per subnet with a secure auto-generated file system policy (VPC-scoped, account-root principal):
+
+```hcl
+module "s3_files" {
+  source = "terraform-aws-modules/s3-bucket/aws//modules/s3-files"
+
+  s3_uri   = "arn:aws:s3:::my-versioned-bucket"
```
antonbabenko: Probably not "s3_uri" but "bucket_arn". Review and fix everywhere.

modules/s3-files/main.tf
```diff
@@ -0,0 +1,169 @@
...
+  subnet_ids = var.subnet_ids
+}
+
+# The central S3 Files file system resource.  A single file system is created
+# per invocation of this module; mount targets (one per subnet) are managed
+# separately below.
+resource "aws_s3files_file_system" "this" {
+  count = var.create ? 1 : 0
+
+  # Pass through the provider region only when the caller explicitly sets it;
```
antonbabenko: Not specific to this module, so please drop such comments.

antonbabenko (changes requested):
Mostly comments, but also a few important fixes required before merge.

sharansutrapu:
Thanks for the review, @antonbabenko ! 

I have applied all the requested changes:
* Dropped all the noisy explanatory comments across `main.tf`.
* Removed the useless `locals` block and mapped directly to `var.subnet_ids`.
* Fixed the lingering `s3_uri` reference to `bucket_arn` in the submodule README.
* Stripped the redundant sections and validation strings from the example README.
* Reordered `examples/s3-files/main.tf` so the core module calls sit cleanly at the top.

The branch is also rebased against the latest `master`. Let me know if everything looks good to merge!

modules/s3-files/main.tf
antonbabenko: This for_each fix is still relevant. Please fix it before merge.

sharansutrapu: I completely agree that `for_each` is the best practice for stable state, so I went ahead and refactored the mount targets to use `for_each = toset(var.subnet_ids)`.

However, testing the `examples/s3-files` directory immediately hits the classic Terraform `(known after apply)` limitation. Because the example provisions the VPC and the S3 Files module in the exact same state file, Terraform blocks the plan since it cannot use unknown subnet IDs as loop keys.

How would you prefer to handle this to ensure the example still works seamlessly out-of-the-box?

1. Change the variable to a map: We could update `var.subnet_ids` to accept a `map(string)` so callers can provide static, known keys for the loop (e.g., `subnet_a = module.vpc.private_subnets[0]`).
2. Revert back to `count`: We could revert just this one resource back to `count` as a pragmatic fix to ensure Day-1 deployments don't require a 2-step targeted apply.

Let me know which approach aligns best with your vision for the module and I will get it pushed up immediately!

antonbabenko: @bryantbiggs could you please review this for_each case and remind me how we handle similar situations in other modules now? Were there any improvements in recent versions of terraform related to "known after apply" issues?

sharansutrapu: Hey @antonbabenko  and @bryantbiggs

I know Anton pinged Bryant on this, but I wanted to help out and did some quick digging into how other modules in the org handle this same `for_each` edge case. 

Looking at `terraform-aws-efs` (which has an identical requirement for per-subnet mount targets), it handles the `(known after apply)` limitation by forcing the variable to be a map rather than a list. 

By doing this, the `examples/complete/main.tf` can dynamically construct a map where the keys are known at plan-time (like AZ names) zipped with the unknown subnet IDs:

mount_targets = { 
  for k, v in zipmap(local.azs, module.vpc.private_subnets) : k => { subnet_id = v } 
}

This perfectly satisfies the for_each DAG requirement while keeping Day-1 deployments seamless!

If this looks like the right path to both of you, I can go ahead and refactor var.subnet_ids in this S3 Files submodule to accept a map(string) and update the example directory to use the zipmap pattern. Just give me the green light and I'll push it up!

examples/s3-files/main.tf
```diff
@@ -0,0 +1,174 @@
...
+        }
+      }
+    }
+  }
+
+  tags = local.tags
+
+  depends_on = [
+    aws_iam_role_policy.s3_files_bucket_access,
+    module.s3_bucket
```
bryantbiggs, line 76: this is not acceptable - this means it should be in the root module and not a sub-module

modules/s3-files/main.tf
bryantbiggs: ya this all just poorly generated by an LLM. its not the right approach - we cannot tell users to add an explicit `depends_on` on another module. this goes directly against Hashi's docs. will try to find time for a proper PR for this

bryantbiggs (changes requested):

sassdavid:
Hi @sharansutrapu,

Do you think you’ll have time to finish this PR? This feature would be very useful for us as well. Would you be able to finish it, or do you need any help with completing it?

sassdavid:
@bryantbiggs before I write more code here, which direction do you want?

I agree callers shouldn't need `depends_on`. But a submodule can't avoid it. `module.s3_bucket.s3_bucket_arn` resolves to `aws_s3_bucket.this`, never to `aws_s3_bucket_versioning.this`, so on destroy the versioning suspend races the file system delete and fails with `BucketHasS3FileSystemAttached`. In the root module it's one line, and `aws_s3_bucket_lifecycle_configuration` already does it:

```hcl
depends_on = [
  # Must have bucket versioning enabled first
  aws_s3_bucket_versioning.this,
```

So:

1. Move `aws_s3files_*` into the root module behind `create_file_system`. No caller `depends_on`. Adds about ten root variables, and the file system only works on buckets this module manages.
2. Keep the submodule so it also works on buckets managed elsewhere.

If you're already writing this yourself, say so and I'll stay out of the way. Otherwise I have the `for_each` fix ready plus: `posix_user` is optional in `CreateAccessPointInput` but required in the PR, the README says `>= 6.39` (should be 6.41+) and `ip_address_type = "ipv4"` (valid values are IPV4_ONLY, IPV6_ONLY, DUAL_STACK).

sharansutrapu:
Hi @sassdavid  and @bryantbiggs 

The current submodule implementation was built to align with the initial discussions on the issue, and it does work safely for isolated use cases. However, Bryant makes a completely fair point regarding the depends_on anti-pattern-moving this into the root module is definitely the right architectural call to handle the versioning lifecycle natively.

When Bryant mentioned he might try to find time to open a proper PR for this, I held off on pushing further updates so I wouldn't step on his toes. That said, I have actually already written the root-level implementation locally to address this exact race condition.

If it saves everyone time, I am more than happy to push that root-level code here for review so we don't have to start from scratch. Just let me know if you'd like me to push it up!

bryantbiggs:
I am working through the backlog on some core changes to make future PRs easier - #408

it has to be balanced with breaking changes. trying to get as much in on the current module major version before we have to a major version bump for longstanding issues that are breaking that we need to make

sharansutrapu:
Totally get it, @bryantbiggs juggling core refactors alongside a major version bump is a lot, so prioritizing #408 makes complete sense.

I don't want to add merge conflicts or extra noise while you're deep in that. I've already got a working root-level implementation for the S3 Files integration on my end, so happy to just paste the code snippets here as a reference for you or David to grab whenever the time is right, or I can just sit tight until #408 settles down. Whatever's easier on your end, I'm good either way!

PR #397, opened 2026-06-05
fix: Wildcard region in WAF/Log delivery policy to support CloudFront

### LINKED ISSUE #396: attach_waf_log_delivery_policy rejects CloudFront standard logging v2 deliveries when the destination bucket is not in us-east-1
The `attach_waf_log_delivery_policy` helper hardcodes `aws:SourceArn` to the bucket's own region via `data.aws_region.current.region`:
```
  condition {
    test     = "ArnLike"
    values   = ["arn:${data.aws_partition.current.partition}:logs:${data.aws_region.current.region}:${data.aws_caller_identity.current.id}:*"]
    variable = "aws:SourceArn"
  }
```

This works for same-region vended-log sources (WAF, ALB, NLB, AppSync) because their CloudWatch Logs delivery resources land in the bucket's region.
  
It does not work for **CloudFront standard logging v2 —** CloudFront is a global service and AWS places its delivery resources in us-east-1 (the CloudFront control plane region) regardless of where the destination S3 bucket lives. When the destination bucket is in, e.g., us-east-2, the policy emits:
  
`"ArnLike": { "aws:SourceArn": "arn:aws:logs:us-east-2:<account>:*" }`
  
 …but the actual delivery destination ARN is:
  
 `arn:aws:logs:us-east-1:<account>:delivery-destination:<name>`

 The region mismatch causes the `s3:PutObject` to be silently denied by the bucket policy, and the destination bucket stays empty.
  
Reproduction

1. Create an S3 bucket via this module in any region other than us-east-1, with `attach_waf_log_delivery_policy = true`.
2. Create a CloudFront distribution and configure CloudFront standard logging v2 → S3 (`aws_cloudwatch_log_delivery_source/_destination/_delivery`) pointing at that bucket.
3. Generate traffic on the distribution; wait the buffer interval (~5–30 min).
4. Observed: bucket remains empty. **CloudWatch Logs DeliveryErrorLogEvents** metric shows access-denied errors against the bucket.
  
Workaround we've used: switch to `attach_lb_log_delivery_policy = true` (same `delivery.logs.amazonaws.com` principal, no region-scoped `SourceArn` condition). Works for both CloudFront (cross-region) and ALB/NLB (same-region), at the cost of broader resource scope (`${bucket}/*`) and losing the `SourceAccount/SourceArn` defense-in-depth on the WAF variant.

sharansutrapu:
## Description
This PR resolves an issue where the `attach_waf_log_delivery_policy` inadvertently blocked log delivery from global AWS services (like CloudFront standard logging v2).

Because CloudFront operates its control plane and log delivery out of `us-east-1`, attempting to log to a bucket in any other region results in an `Access Denied` error. The bucket policy was strictly enforcing that the `aws:SourceArn` match the bucket's local region via `${data.aws_region.current.region}`. 

By replacing the region segment of the ARN with a wildcard (`*`), this PR allows cross-region log delivery from global services while securely maintaining the `${data.aws_caller_identity.current.id}` restriction to prevent cross-account spoofing. I updated both the `WafLogDeliveryWrite` and `WafLogDeliveryAclCheck` statements.

## Motivation and Context
Closes #396. Restores functionality for CloudFront standard logging v2 while maintaining defense-in-depth on the delivery principal.

## Breaking Changes
None. This slightly relaxes the region constraint of the policy but maintains strict account-level boundaries.

## How Has This Been Tested?
- [x] Verified `terraform validate` passes cleanly.
- [x] Executed a full `terraform plan`, `apply`, and `destroy` cycle against a live AWS environment using `examples/complete`. Confirmed the AWS API successfully parses and accepts the wildcarded `aws:SourceArn` within the generated bucket policy.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -910,7 +910,7 @@ data "aws_iam_policy_document" "waf_log_delivery" {
 
     condition {
       test     = "ArnLike"
-      values   = ["arn:${data.aws_partition.current.partition}:logs:${data.aws_region.current.region}:${data.aws_caller_identity.current.id}:*"]
+      values   = ["arn:${data.aws_partition.current.partition}:logs:*:${data.aws_caller_identity.current.id}:*"]
       variable = "aws:SourceArn"
     }
   }
@@ -941,7 +941,7 @@ data "aws_iam_policy_document" "waf_log_delivery" {
 
     condition {
       test     = "ArnLike"
-      values   = ["arn:${data.aws_partition.current.partition}:logs:${data.aws_region.current.region}:${data.aws_caller_identity.current.id}:*"]
+      values   = ["arn:${data.aws_partition.current.partition}:logs:*:${data.aws_caller_identity.current.id}:*"]
       variable = "aws:SourceArn"
     }
   }


sharansutrapu:
Rebased against the latest master to keep everything up to date and removing the stale label. All checks are passing cleanly. Let me know if there is anything else needed from my side for review!

sharansutrapu:
Hi @milosveljkovic and @magreenbaum  gentle bump on this one whenever you have a cycle! 

This is a very small, 2-line fix replacing the region segment with a wildcard to unblock CloudFront log delivery (resolving #396), while still keeping the strict account-level boundary intact. 

It is fully rebased against `master` and all 20 CI checks are completely green. Let me know if you need any tweaks or if it's good to go!

bryantbiggs (approved):

antonbabenko:
This PR is included in version 5.15.3 :tada:

PR #406, opened 2026-08-07
feat(#405): Remove legacy ELB access log delivery policy from main.tf

### LINKED ISSUE #405: Remove legacy ELB access log delivery policy
## Is your request related to a new offering from AWS?

No. This request is related to an existing AWS feature (Application/Network Load Balancer access logging to Amazon S3) that is already supported by AWS and the Terraform AWS Provider.

AWS now recommends granting access log delivery permissions through the `logdelivery.elasticloadbalancing.amazonaws.com` service principal. The previous account-based permissions model is still supported but documented as a legacy approach.

The S3 module currently generates bucket policies containing both the recommended service principal and the legacy account-based permissions. As a result, security scanners and compliance tools may raise alerts because legacy permissions are still present in the bucket policy.

There is currently no way to generate a policy that only contains the recommended AWS permissions model.

Remove the legacy ELB log delivery policy statements entirely from the module and rely only on the currently recommended AWS service principal mechanism.

This would allow users to align with current AWS recommendations and avoid compliance or security alerts triggered by the presence of legacy permissions.

The current workaround is to stop using the module-generated policy and manage the bucket policy manually.

However, this increases maintenance overhead and defeats part of the purpose of using the module.

## Additional context

According to the AWS documentation, the service principal `logdelivery.elasticloadbalancing.amazonaws.com` is the recommended approach, while the account-based permissions model is maintained for backward compatibility.

AWS documentation:
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html

If complete removal of the legacy policy is considered a breaking change, providing a way to disable it would still allow users to adopt the AWS-recommended configuration.

otmane-bouchtaoui-pro:
## Description

Remove legacy Elastic Load Balancing access log delivery bucket policy statements and rely exclusively on the AWS-recommended service principal `logdelivery.elasticloadbalancing.amazonaws.com`.

The module currently generates both the recommended service principal permissions and legacy account-based permissions. This change removes the legacy policy statements while preserving the existing access logging functionality.

## Motivation and Context

AWS recommends granting access log delivery permissions through the `logdelivery.elasticloadbalancing.amazonaws.com` service principal.

The additional legacy account-based permissions are no longer necessary and can trigger findings in security and compliance scanning tools because they grant access using a deprecated permissions model.

Removing these statements aligns the generated bucket policy with current AWS recommendations and the principle of least privilege.

Fixes #405

## Breaking Changes

No.

This change does not alter the module interface or require any configuration changes from users.

AWS Elastic Load Balancing access log delivery continues to function as before using the AWS-recommended service principal `logdelivery.elasticloadbalancing.amazonaws.com`.

The change only removes legacy account-based policy statements that are retained by AWS for backward compatibility and are no longer required when using the recommended service principal model.

## How Has This Been Tested?

  - Verified that the generated bucket policy only contains the `logdelivery.elasticloadbalancing.amazonaws.com` service principal permissions
  - Verified that Terraform plans are generated successfully
  - Verified that legacy ELB account-based policy statements are no longer present
  - Verified that Application Load Balancer access logs continue to be successfully delivered to the S3 bucket

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,6 @@ No modules.
 | [aws_iam_policy_document.require_latest_tls](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) | data source |
 | [aws_iam_policy_document.waf_log_delivery](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) | data source |
 | [aws_partition.current](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/partition) | data source |
-| [aws_region.current](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/region) | data source |
 
 ## Inputs
 
--- a/main.tf
+++ b/main.tf
@@ -1,6 +1,3 @@
-data "aws_region" "current" {
-  region = var.region
-}
 
 data "aws_canonical_user_id" "this" {
   count = local.create_bucket && local.create_bucket_acl && try(var.owner["id"], null) == null ? 1 : 0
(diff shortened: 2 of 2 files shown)


eugenep-scribd:
@bryantbiggs anything prevents this from being merged? 

Amazon does recommend using new policy [docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/enable-access-logging.html#attach-bucket-policy):

> Previously, for Regions available before August 2022, we required a policy that granted permissions to an Elastic Load Balancing account that was specific to the Region. This legacy policy is still supported, but **we recommend that you replace it with the newer policy above**. If you prefer to keep using the legacy policy, which is not shown here, you can.

(I added emphasis)

bryantbiggs:
its a breaking change so it will have to be merged into the next breaking change

PR #408, opened 2026-08-27
feat: Type variable attributes against the provider schema

bryantbiggs:
## Description

- Type every variable against the provider schema; `type = any` is now gone from the module
- Remove the Terragrunt `try(jsondecode(...))` escape hatch, which only existed because of `any`
- Convert every `try()` reading a now-optional attribute. `try` does not catch null, so
  `try(x, default)` silently yields null once an attribute is `optional()`
- Outputs fall back to `null` rather than `""`. This also fixes
  `s3_bucket_lifecycle_configuration_rules`, a list-valued output whose fallback was a string
- `logging` default moves from `{}` to `null`, the only default change. With a `{}` default
  there is no plan-known way to test whether the caller set it, so the `count` guard fails
  when `target_bucket` is a computed value
- `metric_configuration` stays a list rather than the keyed map house style wants. Moving to a
  map re-addresses and recreates existing configurations, so it is left for the next major
- Descriptions follow the provider's wording, wrap literal values in backticks, and drop the
  `Terraform 0.11.12` reference
- `TODO(v6)` markers on what cannot be fixed without a major: `create_bucket`, `versioning`,
  `metric_configuration`, `acl`, the output prefixes, and `attach_public_policy`, whose name
  and description both describe a policy while it gates the public access block

### Examples

- `examples/complete` runs on `BucketOwnerEnforced`, the AWS default since April 2023, which
  takes it from 482 lines to 339. ACL configuration moves to a dedicated `acl` root
- New `bucket-policies` root for the `attach_*_policy` surface
- Fix `examples/directory-bucket`, which fails to apply on `master`. It selects `zone_ids[1]`,
  and S3 Express One Zone is only offered in a subset of each region's AZs
- `s3-analytics` and `s3-inventory` merge into `inventory-and-analytics`; the two were
  separate roots demonstrating two arguments of the same shape
- Drop the `random` provider, `bucket_prefix` already covers global uniqueness
- Raw `aws_kms_key` resources replaced with `terraform-aws-modules/kms/aws`, and our own
  modules pinned to their latest majors, which surfaced two `sns` v7 interface changes
- 70 of 72 root variables and all 65 sub-module variables are now exercised. The two that are
  not are `attach_public_policy` and `skip_destroy_public_access_block`; covering either means
  an example that disables a public access block, which is not what these should teach
- Every example root now reads metadata, then the module invocations it exists to
  demonstrate, then supporting infrastructure under `# Supporting Resources`, with the
  `create = false` invocation last in the middle section. In `examples/notification` the
  module under test had drifted beneath a `# Lambda Function Package` banner that described
  the block above it
- Example READMEs all carry the same `## Usage` block and cost note; two did not

### Also

- `examples/account-public-access` required Terraform `>= 1.10` while every other directory in
  the repo requires `>= 1.5.7`. The commit that lowered the floor missed this one file, so CI
  has been testing that root at the wrong minimum since June 2025
- Two dead links: `docs/UPGRADE-3.0.md` pointed at `examples/complete-legacy` on `master`,
  deleted by that same commit, and now points at `v5.0.0`; the S3 Vectors product page moved
  to `aws.amazon.com/s3/features/vectors/`

## Motivation and Context

- `type = any` gives callers no signal about what a variable accepts, defers every shape error
  to apply, and publishes nothing useful into the README
- It also forced the `jsondecode` workaround and a `map(string)` interface for `versioning`
  that had to be converted back with a four-deep `try` chain

## Breaking Changes

- No. Nothing is renamed, no version floor moves, no resource address changes. Terraform drops
  unknown attributes when converting to an object type, so callers passing extra keys are
  unaffected, as they were under `any`
- The `logging` default change affects nobody observed: of the public callers that set
  `logging`, all 26 pass a populated object and none passes `{}` or `null`

## How Has This Been Tested?

- Every example applied, re-planned clean, and destroyed. `validate` alone missed seven
  defects, including three `count` guards that became dependent on computed values
- Upgrade tested: each example deployed from `master`, then the module swapped underneath it
  keeping master's config and state. `No changes`, or output-only diffs Terraform describes as
  "without changing any real infrastructure"


magreenbaum (approved):
Looks good, thank you!

bryantbiggs:
cc @antonbabenko if you get a moment

antonbabenko (approved):
Amazing! Sorry for the delay with the review!

antonbabenko:
This PR is included in version 5.16.0 :tada:

AdrienKuhn:
Causing breaking changes https://github.com/terraform-aws-modules/terraform-aws-s3-bucket/issues/409

PR #410, opened 2026-09-15
feat: Add Amazon S3 Files support

### LINKED ISSUE #386: Proposal: add an Amazon S3 Files submodule when provider support is available

bryantbiggs:
## Description

Adds [Amazon S3 Files](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-files.html) support to the root module through a new `file_systems` map. Each file system can be scoped to a prefix and gets:

- an IAM role scoped to its prefix, or your own via `iam_role_arn`
- mount targets, on one shared security group by default or the file system's own `security_groups`
- access points, where `read_access_arns` / `read_write_access_arns` generate the allow and deny statements AWS recommends
- an optional file system policy and synchronization configuration

File systems are not created on a directory bucket, which S3 Files does not support. New example: `examples/file-system`.

Also:

- README usage restructured: one log delivery snippet instead of three, and the feature list split into capabilities and bucket types
- `attach_elb_log_delivery_policy` and `attach_lb_log_delivery_policy` descriptions now name the load balancers each one serves
- Fixes an existing bug in `modules/table-bucket`, unrelated to S3 Files: a `table_bucket_policy_statements` entry with `not_resources` also received the default `resources`, and IAM rejects a statement with both

## Motivation and Context

- Resolves #386
- Closes #388 

It lives in the root module so the file system is destroyed before bucket versioning is suspended, without callers adding `depends_on`.

## Breaking Changes

None. The minimum AWS provider version moves from 6.42 to 6.44, which fixes deleting a synchronization configuration (hashicorp/terraform-provider-aws#47753).

## How Has This Been Tested?

Against a live account: `examples/file-system` applied, re-planned with no changes and destroyed; a create, update, replace, bring-your-own and delete walk; a client mounting through an access point; and the upgrade test on `examples/complete` with `No changes`.


sassdavid:
Hi @bryantbiggs,

Thanks for putting this together, it covers a lot of ground. I went through the branch and two things came up that you may want to look at. Both are questions rather than requests.

**1. `create_policy = true` with only `source_policy_documents`**

The use case is attaching a policy document that is already built elsewhere in the config:

```hcl
file_systems = {
  example = {
    create_policy           = true
    source_policy_documents = [data.aws_iam_policy_document.shared.json]
  }
}
```

No `policy_statements`, and no access point names principals, so `data.aws_iam_policy_document.policy` ends up with zero `statement` blocks. That attribute is then `null` rather than an empty list, so the precondition at `modules/file-system/main.tf:592` fails before its own message is reached:

```
Error: Invalid function argument
  while calling length(value)
  data.aws_iam_policy_document.policy[0].statement is null
```

Statements from `source_policy_documents` also land only in `.json`, never in `.statement`, so the count stays 0 even past the null. Would `try(length(...), 0)` plus a look at the source documents fit here, or did you have a different shape in mind?

**2. The versioning precondition message**

`bucket_versioning_status` comes from `aws_s3_bucket_versioning.this[0]`, which only exists when `var.versioning` is non-empty. On a bucket where versioning is already enabled but managed outside this module, the input is `null` and the plan stops with:

```
S3 Files requires versioning to be enabled on the bucket.
```

Versioning is enabled, so the message sends you looking in the wrong place. `versioning = { enabled = true }` sorts it out. Would it be worth saying that the module has to manage versioning for the ordering to work?

Two smaller notes, no strong feelings:

- `kms_key_id` is `fwtypes.ARNType` in the provider, so a key ID is rejected with `Invalid ARN Value`. The description reads "ID of the KMS key".
- `iam_role_policies` is filtered by `if local.create_iam_role`, so it is dropped quietly when a role is brought in.

